PR #1, opened 2025-01-13
setup monorepo infrastructure

bajtos:
Directories:

```
api/
  Public REST API exposing info about FIL+ deals

backend/
  Private singleton observing built-in actor events and maintaining the database of FIL+ deals

db/
  Shared component providing helpers to access the database, including database schema migration scripts.
```

Features:
- linting: prettier + standard
- TypeScript checks
- tests: node:test + node:assert
- CI/CD
- deployment to Fly.io

Code was created by combining setup from  the following monorepos:
- https://github.com/filecoin-station/spark-stats
- https://github.com/filecoin-station/piece-indexer

Links:

- https://github.com/space-meridian/roadmap/issues/172

## TODO

- [x] Create Sentry project(s) and configure DSNs
- [x] Create InfluxDB bucket and access token, configure the token in Fly.io secrets
- [x] Create Fly.io apps
- [x] Create Fly.io deployment API tokens and update GHA workflows

## FAQ

**Why Postgres and not Redis?**

In spark-api, we are providing REST API for summarising the eligible deals per miner/client/allocator, see [the routes here](https://github.com/filecoin-station/spark-api/blob/defd16cca24ec1ff1ac7514f6f84286e00c68cbf/api/index.js#L31-L36).

If we used Redis, we would have to implement custom indices or clever queries to aggregate the deals. Using Postgres' built-in indexing & querying seems like a much easier way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,3 +15,16 @@ of FIL+ deals
 
 Shared component providing helpers to access the database, including database
 schema migration scripts.
+
+## Deployment
+
+All commits pushed to the `main` branch are automatically deployed to Fly.io by
+a GitHub Actions workflow.
+
+To deploy the changes manually, run the following command from the **monorepo
+root directory**:
+
+```bash
+fly deploy -c api/fly.toml
+fly deploy -c backend/fly.toml
+```
--- a/backend/fly.toml
+++ b/backend/fly.toml
@@ -1,4 +1,4 @@
-app = "spark-observer-backend"
+app = "deal-observer-backend"
 primary_region = "cdg"
 
 [env]


juliangruber:
nit: `setup` is not a verb, should be `to set up`

.github/dependabot.yml
```diff
@@ -0,0 +1,39 @@
...
+    directories:
+      - "/"
+    schedule:
+      interval: "daily"
+      time: "15:00"
+      timezone: "Europe/Berlin"
+    reviewers:
+      - "bajtos"
+      - "juliangruber"
+      - "NikolasHaimerl"
```
juliangruber, line 39: https://github.com/filecoin-station/deal-observer/pull/14

.github/workflows/ci.yml
```diff
@@ -0,0 +1,157 @@
+name: CI
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
```
juliangruber, line 6: https://github.com/filecoin-station/deal-observer/pull/15

README.md
```diff
@@ -1,3 +1,30 @@
 # deal-observer
 
 A near-real-time database of active FIL+ deals
+
+### [api/](./api/)
+
+Public REST API exposing info about FIL+ deals.
+
+### [backend/](./backend/)
```
juliangruber, line 9: "backend" is slightly confusing, as it's not clear where the api ends and the backend begins. Would "observer" or "ingestion" be clearer?

backend/bin/deal-observer-backend.js
```diff
@@ -0,0 +1,54 @@
...
+
+const pgPool = await createPgPool()
+
+const fetchRequest = new ethers.FetchRequest(RPC_URL)
+fetchRequest.setHeader('Authorization', rpcHeaders.Authorization || '')
+const provider = new ethers.JsonRpcProvider(fetchRequest, null, { polling: true })
+
+const { recordTelemetry } = createInflux(INFLUXDB_TOKEN)
+
+const loop = async (name, fn, interval) => {
```
juliangruber, line 25: At this stage of the project, the `loop` function is overkill. It was obviously designed for the case where you have multiple loops. I don't want to cause any extra work though, if you don't want to refactor this could you please add a comment why this function exists?

backend/lib/config.js
```diff
@@ -0,0 +1,18 @@
+const {
+  RPC_URLS = 'https://api.node.glif.io/rpc/v0',
```
juliangruber, line 2: https://github.com/filecoin-station/deal-observer/pull/16

juliangruber (commented):
👏

README.md
bajtos: We use the name "indexer" in piece-indexer and "observer" in "spark-observer". I find it weird to have "spark-observer-observer" component, that's why I used "deal-observer-backend" here.

I see how "backend" may be slightly confusing, though.

Can we find a better name conveying the information that this is the backend (hidden) service observing the deals in the background and updating the database?

backend/bin/deal-observer-backend.js
bajtos: Very soon, we will need to add a second loop to add Payload CID to deals ingested from DDO events. Then we will need a third loop to sync DDO-sources deals to spark-api.

Having said that, @NikolasHaimerl is removing the `loop` helper in #12, so I think this comment will be resolved soon.

README.md
juliangruber: We also have `api` in `spark-api` though, do you find it confusing there?

PR #8, opened 2025-01-15
feat: poc retrieve ddo claims transformer

NikolasHaimerl:

1. Add a transformer which handles the parsing of built in actor events. 

Context: We are using the [glif api](https://api.node.glif.io/) to retrieve data from Filecoin. In Filecoin built in [actors](https://docs.filecoin.io/reference/built-in-actors) are responsible for managing and updating the global state of the filecoin network. 
DDO (Direct Data Oboarding) in FIL+ deals bypass the usually used StorageActor and oboard data directly. The actor which registers the oboarded data is the verified registry actor ([Source](https://github.com/filecoin-project/FIPs/discussions/730)). 
For spark to be aware of these deals we need to listen to the events emmitted by this actor and look for deals with certain properties (new deals, updated deals,...). 

The glif api allows for the [fetching](https://playground.open-rpc.org/?url=https://api.node.glif.io) of raw built-in Actorevents. The returned data needs to be processed, extracted and then parsed into the correct actor event type. 
In this PR the parsing of the ClaimEvent type is being implemented. A sample data point of how the fetched data (after some preprocessing) might look like is stored in `test_data`.  Here, we test that the data can be successfully parsed by the transformer and a class object of the type ClaimEvent can be retrieved for further use. 

This implementation is inspired by an existing [prototype](https://github.com/rvagg/spacemon) for fetching build in actor events. 

See [issue](https://github.com/space-meridian/roadmap/issues/172).


backend/test/test_data/claim_event.json
```diff
@@ -0,0 +1,13 @@
...
+    "provider": 3254061,
+    "pieceCid": {
+        "/": "baga6ea4seaqm3i2rjffnsz55jvzod52yprjmxco3x2wzjocjxmqc7y7q3nxrena"
+    },
+    "pieceSize": 34359738368,
+    "termMin": 1008000,
+    "termMax": 1267200,
+    "termStart": 4620803,
+    "sector": 102905
+}
```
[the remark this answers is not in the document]
NikolasHaimerl: I think it is much easier to read if we have the type definitions collected in an ipldsch file. This will be especifally useful once we start adding more Event types to the list of events we want to listen to. Having all the json object extraction inside of a javascript file adds a lot of boilerplate code that can be hidden away using type parsing with ipld.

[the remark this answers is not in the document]
bajtos: Fair enough 👍🏻

backend/lib/lotus/transform.js
```diff
@@ -0,0 +1,39 @@
...
+import { fromDSL } from '@ipld/schema/from-dsl.js'
+import { create } from '@ipld/schema/typed.js'
+
+class Transformer {
+  // A transformer which takes in a json object and returns a typed ClaimEvent object
+  #claimEventTransformer
+
+  async build () {
+    const schemaPath = new URL('../lotus/builtin-actor-events-schemas.ipldsch', import.meta.url)
+    // TODO: Catch and log errors
```
bajtos, line 11: If you don't mind, let's ignore error handling for now and focus on getting a full vertical slice completed first, including the loop observing the events via the RPC API and updating the database.

Then we can clean up the rough edges and make the implementation more robust.

Leaving TODO comments like this one is a great solution for this 👍🏻

NikolasHaimerl: Yes, I was thinking of adding rigorous logging for the server at some point. But implementing logging when the server itselve is not yet up seemed a bit premature for me. I just wanted to give myselve a reminder where possible errors could spawn.

backend/test/observer.test.js
```diff
@@ -32,4 +37,33 @@ describe('deal-observer-backend', () => {
...
+          }
+        })
+      })
+    })
+
+    it('transformer can correctly transform claim event json objects', async () => {
+      const transformer = await (new Transformer().build())
+      const claimEvent = testData.claimEvent
+      const transformedClaimEvent = transformer.transform('ClaimEvent', claimEvent)
+      assert.deepStrictEqual(transformedClaimEvent, testData.typedClaimEvent)
```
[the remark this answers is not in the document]
NikolasHaimerl: Thanks for catching that, it was a typo

backend/lib/lotus/builtin-actor-events-schemas.ipldsch
```diff
@@ -0,0 +1,24 @@
...
+  client Int
+  provider Int
+  pieceCid &Any 
+  pieceSize Int 
+  termMin Int 
+  termMax Int 
+  termStart Int 
+  sector Int 
+}
+
```
pyropy, line 12: We could potentially run `ipld-schema validate` on this file as part of our CI

backend/lib/lotus/transform.js
```diff
@@ -0,0 +1,39 @@
+import { readFile } from 'node:fs/promises'
+import { fromDSL } from '@ipld/schema/from-dsl.js'
+import { create } from '@ipld/schema/typed.js'
+
+class Transformer {
+  // A transformer which takes in a json object and returns a typed ClaimEvent object
+  #claimEventTransformer
+
+  async build () {
+    const schemaPath = new URL('../lotus/builtin-actor-events-schemas.ipldsch', import.meta.url)
```
pyropy, line 10: I wonder if we should convert IPLD schema to JSON and import is at a object instead of importing form a relative path. Seems also like it's possible to convert schema to JS code too.

backend/test/observer.test.js
```diff
@@ -32,4 +38,35 @@ describe('deal-observer-backend', () => {
...
+    const testData = {}
+    beforeEach(async () => {
+      const __filename = fileURLToPath(import.meta.url)
+      const __dirname = path.dirname(__filename)
+
+      const testDataPath = path.join(__dirname, 'test_data')
+      // Read all files in the test_data directory
+      fs.readdir(testDataPath, (_, files) => {
+        files.forEach((file) => {
+          if (path.extname(file) === '.json') {
```
pyropy, line 52: To avoid reading data from json files we could hold test data inside a JS file, similar like [we're doing it inside spark-evaluate.](https://github.com/filecoin-station/spark-evaluate/blob/main/test/helpers/test-data.js)

backend/test/test_data/claimEvent.json
```diff
@@ -0,0 +1,11 @@
+{
```
pyropy, line 1: We could store this value as a variable inside some javascript file, similar to [this](https://github.com/filecoin-station/spark-evaluate/blob/d2368132e8c8987c87cb5700dc13dfeda291cd17/test/helpers/test-data.js#L11-L17).

pyropy (changes requested):
Let's try to move test data to a JavaScript file so we avoid reading the test data files before each test.

backend/lib/lotus/builtin-actor-events-schemas.ipldsch
NikolasHaimerl: Good idea, I'll add it to the workflor

backend/lib/lotus/transform.js
NikolasHaimerl: Thanks for the tip. This makes the code more readable and more concise.

backend/test/observer.test.js
NikolasHaimerl: Great suggestion!

backend/lib/lotus/builtin-actor-events-schemas.ipldsch
NikolasHaimerl: Actually, if we confert the ipldsch file to js and use it during testing we have a ipld-schema testing built in because it has to pass `fromDSL` anyway for the transformer test to succeed. I think this should be succifient. WDYT?

bajtos:
> Let's try to move test data to a JavaScript file so we avoid reading the test data files before each test.

+1

When the test data is loaded from an external files, it's difficult to reason about a test, because it's not clear what are the inputs and expected outputs, one has to look into other files which are cumbersome to look up.

backend/test/observer.test.js
```diff
@@ -32,4 +34,13 @@ describe('deal-observer-backend', () => {
       await observeBuiltinActorEvents(pgPool, providerMock)
     })
   })
+
+  describe('Transformer', () => {
+    it('transforms a claim event payload to a typed object', async () => {
+      const transformer = await (new Transformer().build())
+      const transformedClaimEvent = transformer.transform('ClaimEvent', claimTestEvent)
+      assert(transformedClaimEvent !== undefined, 'transformedClaimEvent is undefined')
+      assert.deepStrictEqual(transformedClaimEvent, claimTestEvent)
```
bajtos, line 43: I find this super surprising. If the transformed event is the same object as the original event, then what's the point of the transformation step?

We are missing key information here: what is the payload returned by the JSON-RPC API?

bajtos (changes requested):

backend/lib/lotus/builtin-actor-events-schemas.ipldsch
pyropy: If that's the case I guess it will be okay for now. Schema is still sufficiently small and loading schema in that way should be maintainable. 👍🏻

backend/test/test_data/claimEvent.js
```diff
@@ -0,0 +1,15 @@
...
+  client: 3297188,
+  provider: 3254061,
+  pieceCid: CID.parse('baga6ea4seaqk5bmtyfomjprov2jjtto7ffhxuusfouhn3fdu7l6gl6yemhzocoi'),
+  pieceSize: 34359738368,
+  termMin: 1008000,
+  termMax: 1267200,
+  termStart: 4620803,
+  sector: 102905
+}
+
```
pyropy, line 14:
```suggestion
Object.freeze(claimTestEvent);
```

We can also freeze test objects to make sure data is not mutated in the tests.

backend/lib/lotus/transform.js
```diff
@@ -0,0 +1,32 @@
+import { create } from '@ipld/schema/typed.js'
+import { schemaDmt } from './builtin-actor-events-schemas.js'
+
+class Transformer {
```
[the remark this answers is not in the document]
bajtos: > Proposed names: `ActorEventDeserializer` or `EventPayloadParser`

☝🏻 that's just for inspiration, feel free to use a different name, one that works best for you.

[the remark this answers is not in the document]
NikolasHaimerl: The functionality of the transformer is not converting something form a serialized format into a deserialized format but rather check the type of a json object and guarentee a typed class instance if the transformation is successful. 

I could call it IpldSchema and the corresponding function `parseSchema`. That would be an accurate representation of what the transformer is doing.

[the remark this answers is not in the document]
pyropy: @NikolasHaimerl so functionality of the transformer is to validate json schema of the supplied data? Maybe we should go with something like `EventPayloadValidator` and instead of `transform` we could have a `validate` method?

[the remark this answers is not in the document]
bajtos: +1 to use "validator" in the class name and "validate" as the method name if this class only validates conformance with the IPLD schema.

juliangruber:
Why did you close this PR @NikolasHaimerl?

PR #12, opened 2025-01-16
feat: poc retrieve ddo claims

NikolasHaimerl:
This MR proposes the following changes 

1. Add an rpc service for retrieving data from glif
2. Add an ilpd schema

Context: We are using the [glif api](https://api.node.glif.io/) to retrieve data from Filecoin. In Filecoin built in [actors](https://docs.filecoin.io/reference/built-in-actors) are responsible for managing and updating the global state of the filecoin network. 
DDO (Direct Data Oboarding) in FIL+ deals bypass the usually used StorageActor and oboard data directly. The actor which registers the oboarded data is the verified registry actor ([Source](https://github.com/filecoin-project/FIPs/discussions/730)). 
For spark to be aware of these deals we need to listen to the events emmitted by this actor and look for deals with certain properties (new deals, updated deals,...). 

The glif api allows for the [fetching](https://playground.open-rpc.org/?url=https://api.node.glif.io) of raw built-in Actorevents. The returned data needs to be processed, extracted and then parsed into the correct actor event type. 
In this PR the parsing of the ClaimEvent type is being implemented. A sample data point of how the fetched data (after some preprocessing) might look like is stored in `test_data`.  Here, we test that the data can be successfully parsed by the transformer and a class object of the type ClaimEvent can be retrieved for further use. 

This implementation is inspired by an existing [prototype](https://github.com/rvagg/spacemon) for fetching build in actor events. 

See [issue](https://github.com/space-meridian/roadmap/issues/172).


backend/lib/config.js
```diff
@@ -4,15 +4,19 @@ const {
 } = process.env
 
 const rpcUrls = RPC_URLS.split(',')
-const RPC_URL = rpcUrls[Math.floor(Math.random() * rpcUrls.length)]
-console.log(`Selected JSON-RPC endpoint ${RPC_URL}`)
+const GLIF_RPC = rpcUrls[Math.floor(Math.random() * rpcUrls.length)]
```
bajtos, line 7: What is the rationale for tying the variable name to Glif? Do you require the JSON-RPC provider to offer any Glif-specific functionality? I believe we should be able to use any Lotus-compatible JSON-RPC API provider like Beryx/Zondax instead of Glif, e.g. as a backup if Glif has an outage.

The variable name should describe the what it can be used for (JSON-RPC provider URL), not implementation/configuration details (the fact that we are using Glif by default).

backend/lib/rpc-service/service.js
```diff
@@ -0,0 +1,92 @@
...
+    headers: { 'content-type': 'application/json' },
+    body: reqBody
+  })
+  return jsonDecode(new Uint8Array(await response.body.arrayBuffer())).result
+}
+
+/*
+ A class to interact with the Lotus HTTP RPC API.
+*/
+class LotusService {
```
bajtos: - We usually use the term "client" instead of "service" for code that provides client-side helpers for accessing a remote server.
- Let's not refer to Lotus please, this should work with any JSON-RPC API provider implementing the standard methods.

Suggested name: `FilecoinRpcApiClient` or something similar (and shorter?).

backend/lib/deal-observer.js
```diff
@@ -1,11 +1,35 @@
...
+
+  constructor (pgPool = null, chainHead = null) {
+    // TODO: Store events in pgPool
+    this.#pgPool = pgPool
+    this.#cache = new Map()
+    this.#cache.set('chainHead', chainHead)
+  }
+
+  async build () {
+    this.#lotusService = await (new LotusService()).build()
```
bajtos: Since an instance of this class cannot be used before it's built, I think it's better to create a static asynchronous factory method.

```suggestion
    this.#lotusService = await LotusService.create()
```

WDYT?

backend/test/utils.js
```diff
@@ -0,0 +1,30 @@
+import { CID } from 'multiformats'
+import { base32 } from 'multiformats/bases/base32'
+
+// Extracting CIDs from JSON objects for easier testing
+const parseCIDs = (jsonObject) => {
```
bajtos, line 5: I believe [@ipld/dag-json](https://www.npmjs.com/package/@ipld/dag-json) already provides this functionality. 

The convention `{ '/': 'CID-string' }` for links is defined in IPLD - see https://ipld.io/docs/codecs/known/dag-json/

Can you use that IPLD package instead of writing a custom deserializer?

backend/test/observer.test.js
```diff
@@ -16,20 +21,59 @@ describe('deal-observer-backend', () => {
...
+      assert(chainHead)
+      assert.deepStrictEqual(JSON.stringify(chainHead), JSON.stringify(chainHeadTestData))
+    })
+
+    it('test the retrieval of rawActorEvents', async () => {
+      const actorEvents = await lotusService.getActorEvents(new ActorEventFilter(4622129, 4622139, ['claim']))
+      assert(actorEvents)
+      actorEvents.forEach(e => {
+        assert(e.height >= 4622129 && e.height <= 4622139)
+      })
```
bajtos, line 76: In JavaScript, assertions cannot see into the parameters. When this check fails, the message will only say "it failed" with no details about which actor event failed and what was the actual height.

We usually use the following approach:

```js
assert.deepStrictEqual(
  actorEvents.map(e => ({ id: e.id, height: e.height })),
  [
    { id: 1, height: 4622129 },
    { id: 2, height: 4622129 },
    { id: 3, height: 4622130 },
  ]
)
```

In your test, the list of events is rather long, so it would be impractical to assert on all of them.

This is IMO a flag that your test is not focused enough and will make it difficult to troubleshoot failures.

A better way to write this test:

- Create a focused `makeRpcRequest` mock that returns a minimal number of events needed to test the behaviour of `getActorEvents`. 
- Then the assertion can assert on the exact event payload.
- If you want to test the event filter, then IMO you should have events 1) just before the start epoch, 2) at the start epoch, 3) in the middle, 4) at the end epoch 5) just after the end epoch.

Based on what I heard from Rodd, some epochs have more events than can be fetched in a single RPC API call. IIUC, the limit is 10k events. 

In that light, it will most likely be much simpler if your service does not accept a range of epochs but only one epoch.

bajtos (changes requested):
The implementation is heading in the right direction in terms of how it fetches & processed the events 👍🏻 

We will need more discussion around the code design and testing, let's do that next week.

In the meantime, I think it would be best if you could continue the work to cover the entire vertical slice, integrating your event fetching & processing into the main observer loop.

backend/test/observer.test.js
NikolasHaimerl: > Based on what I heard from Rodd, some epochs have more events than can be fetched in a single RPC API call. IIUC, the limit is 10k events.
That is a bit unsettling if it is true. I have not seen any information returned by the RPC API on whether the list of events fetched is complete or not. 
Only accepting a single epoch would make the service easier although we would need more calls. I guess since we are switching to the paid version soon this should not be an issue anymore.

backend/lib/deal-observer.js
NikolasHaimerl: Sounds good to me.

backend/lib/config.js
```diff
@@ -4,15 +4,19 @@ const {
...
+console.log(`Selected JSON-RPC endpoint ${GLIF_RPC}`)
 
 const rpcHeaders = {}
-if (RPC_URL.includes('glif')) {
+if (GLIF_RPC.includes('glif')) {
   rpcHeaders.Authorization = `Bearer ${GLIF_TOKEN}`
 }
 
+// The types of events we want to watch out for
+const EVENT_TYPES = ['claim']
```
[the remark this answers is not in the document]
NikolasHaimerl: Fair enough, I'll remove it.

backend/lib/config.js
NikolasHaimerl: I thought it would yield more clarity on which API we are using, but I see that this is not the case. I'll change it back.

backend/test/utils.js
NikolasHaimerl: I'll have a look. It would certainly shorten the implementation.

backend/test/observer.test.js
NikolasHaimerl: I think I could also use the schema validator to validate the correctness of the returned event data. This way I can be sure that all types were correctly passed and that they fit the expected schema.

backend/lib/deal-observer.js
```diff
@@ -1,11 +1,86 @@
...
+          piece_cid,
+          piece_size,
+          term_start_epoch,
+          term_min,
+          term_max,
+          sector_id
+        )
+        VALUES (unnest($1::int[]), unnest($2::int[]), unnest($3::int[]), unnest($4::text[]), unnest($5::bigint[]), unnest($6::int[]), unnest($7::int[]), unnest($8::int[]), unnest($9::bigint[]))
+      `
+    const time2 = Date.now()
```
bajtos, line 65: Since we are no longer looping over deals, we can remove this variable.

backend/lib/deal-observer.js
```diff
@@ -1,11 +1,86 @@
...
+      transformedDeals.map(deal => deal.piece_size),
+      transformedDeals.map(deal => deal.term_start_epoch),
+      transformedDeals.map(deal => deal.term_min),
+      transformedDeals.map(deal => deal.term_max),
+      transformedDeals.map(deal => deal.sector_id)
+    ])
+
+    console.log(`Loop of inserting took ${Date.now() - time2}ms`)
+    // Commit the transaction if all inserts are successful
+    const time1 = Date.now()
```
bajtos, line 80: Ditto, we are no longer looping over deals inside a transaction.

```suggestion
```

backend/lib/deal-observer.js
bajtos:
```suggestion
```

backend/lib/deal-observer.js
```diff
@@ -1,11 +1,86 @@
...
+    console.log(`Inserting ${activeDeals.size} deals took ${time1 - time0}ms`)
+  } catch (error) {
+    // If any error occurs, roll back the transaction
+    console.error('Error inserting deals. Rolling back:', error.message)
```
bajtos, line 84: We are not rolling back any transactions now.

Please update the comment and the log message.

backend/lib/rpc-service/data-types.js
```diff
@@ -0,0 +1,41 @@
+import { Type } from '@sinclair/typebox'
+const ClaimEvent = Type.Object({
+  id: Type.Number(),
+  client: Type.Number(),
+  provider: Type.Number(),
+  pieceCid: Type.Any(),
+  pieceSize: Type.Number(),
```
bajtos, line 7: Please use the BigInt type here to match the type found in the RPC API responses.

backend/lib/rpc-service/service.js
```diff
@@ -0,0 +1,91 @@
...
+import { Value } from '@sinclair/typebox/value'
+import { ClaimEvent, RawActorEvent, BlockEvent } from './data-types.js'
+
+/** @import {CID} from 'multiformats' */
+
+/**
+ * @param {string} method
+ * @param {Object} params
+  * @returns {Promise<object>}
+  */
```
bajtos, line 16: Let's fix the formatting, please.

```suggestion
 * @param {Object} params
 * @returns {Promise<object>}
 */
```

backend/lib/rpc-service/service.js
```diff
@@ -0,0 +1,91 @@
...
+    headers: { 'content-type': 'application/json' },
+    body: reqBody
+  })
+  return jsonDecode(new Uint8Array(await response.body.arrayBuffer())).result
+}
+/**
+     * @param {object} actorEventFilter
+     * Returns actor events filtered by the given actorEventFilter
+     * @returns {Promise<Array<BlockEvent>>}
+     */
```
bajtos, line 32: Ditto.

backend/lib/rpc-service/service.js
```diff
@@ -0,0 +1,91 @@
+import { RPC_URL } from '../config.js'
+import { base64pad } from 'multiformats/bases/base64'
+import { encode as cborEncode } from '@ipld/dag-cbor'
+import { decode as jsonDecode } from '@ipld/dag-json'
+import { request } from 'undici'
```
bajtos, line 5: Let's use the built-in Fetch API instead of undici. It's always better to have less dependencies. Unless you need a specific feature that undici provides but the built-in API does not?

backend/package.json
```diff
@@ -14,11 +14,18 @@
...
+    "@sinclair/typebox": "^0.34.14",
     "@types/slug": "^5.0.9",
     "debug": "^4.4.0",
     "ethers": "^6.13.5",
+    "multiformats": "^13.3.1",
     "pg": "^8.13.1",
-    "slug": "^10.0.0"
+    "request": "^2.88.2",
+    "slug": "^10.0.0",
+    "undici": "^7.2.1"
```
bajtos, line 29: Let's use the built-in Fetch API and remove this dependency.

backend/package.json
```diff
@@ -14,11 +14,18 @@
...
+    "@ipld/schema": "^6.0.5",
     "@sentry/node": "^8.48.0",
+    "@sinclair/typebox": "^0.34.14",
     "@types/slug": "^5.0.9",
     "debug": "^4.4.0",
     "ethers": "^6.13.5",
+    "multiformats": "^13.3.1",
     "pg": "^8.13.1",
-    "slug": "^10.0.0"
+    "request": "^2.88.2",
```
bajtos, line 27: The request module is not maintained anymore. Please use the built-in Fetch API instead.

backend/package.json
```diff
@@ -14,11 +14,18 @@
   "dependencies": {
     "@filecoin-station/deal-observer-db": "^1.0.0",
     "@influxdata/influxdb-client": "^1.35.0",
+    "@ipld/dag-cbor": "^9.2.2",
+    "@ipld/dag-json": "^10.2.3",
+    "@ipld/schema": "^6.0.5",
```
bajtos, line 19: Do we still need `@ipld/schema` now that you are using typebox? If not, then please remove this unused dependency.

backend/test/rpc-client.test.js
```diff
@@ -0,0 +1,43 @@
...
+
+  it('test the retrieval of the chainHead', async () => {
+    const chainHead = await getChainHead(makeRpcRequest)
+    assert(chainHead)
+    const expected = parse(JSON.stringify(chainHeadTestData))
+    assert.deepStrictEqual(chainHead, expected)
+  })
+
+  it('test the retrieval of rawActorEvents', async () => {
+    Array.from({ length: 11 }, (_, i) => 4622129 + i).flatMap(async (blockHeight) => {
```
bajtos, line 30: Let's rewrite this using a for loop please.

To make the assertion errors easier to reason about, let's change this into a parameterized test, so that we have one test case for each block tested.

```js
for (let blockHeight = 4622129; blockHeight < 4622129 + 11; blockHeight++) {
   it(`retrieves rawActorEvents in block ${blockHeight}`, async () => {
     // ...
   })
}
```

backend/test/rpc-client.test.js
```diff
@@ -0,0 +1,43 @@
...
+      const actorEvents = await getActorEvents(getActorEventsFilter(blockHeight, 'claim'), makeRpcRequest)
+      assert(actorEvents)
+      assert(actorEvents.length > 0)
+      actorEvents.forEach(e => {
+        // Validate type
+        const parsedEvent = Value.Parse(ClaimEvent, e.event)
+        assert(parsedEvent, `Invalid claim event: ${JSON.stringify(e.event)}`)
+        assert(e.height >= 4622129 && e.height <= 4622139)
```
bajtos, line 38: IIUC, the event height should match the `blockHeight`.

```suggestion
        assert.strictEqual(e.height, blockHeight)
```

backend/test/rpc-client.test.js
```diff
@@ -0,0 +1,43 @@
...
+      case 'Filecoin.ChainHead':
+        return parse(JSON.stringify(chainHeadTestData))
+      case 'Filecoin.GetActorEventsRaw':
+        return parse(JSON.stringify(rawActorEventTestData)).filter(e => e.height >= params[0].fromHeight && e.height <= params[0].toHeight)
+      default:
+        console.error('Unknown method')
+    }
+  }
+
+  it('test the retrieval of the chainHead', async () => {
```
bajtos, line 22: Let's use the BDD-style test naming. The line should read like an English sentence.

Proposed test name: "RpcApiClient retrieves the chainHead"

`it()` stands for `RpcApiClient` here (the text you have in the `describe` block above).

```suggestion
  it('retrieves the chainHead', async () => {
```

The English sentence:

It _(meaning RpcApiClient)_ retrieves the chainHead.

db/lib/types.js
```diff
@@ -0,0 +1,15 @@
...
+    piece_cid: Type.String(),
+    piece_size: Type.BigInt(),
+    term_start_epoch: Type.Number(),
+    term_min: Type.Number(),
+    term_max: Type.Number(),
+    sector_id: Type.BigInt(),
+    payload_cid: Type.Null()
+})
+
+export { ActiveDealDbEntry }
```
bajtos, line 15: Missing EOL at EOF

db/migrations/002.do.active-deals.sql
```diff
@@ -7,7 +7,7 @@ CREATE TABLE active_deals (
   term_start_epoch INT NOT NULL,
   term_min INT NOT NULL,
   term_max INT NOT NULL,
-  sector_id INT NOT NULL,
+  sector_id BIGINT NOT NULL,
```
bajtos, line 10: You cannot modify existing columns this way, you need to create a new migration script that will `ALTER TABLE active_deals`

bajtos:
@NikolasHaimerl PTAL at the CI results, the linting step is failing.

backend/bin/deal-observer-backend.js
```diff
@@ -1,54 +1,64 @@
...
+// Filecoin will need some epochs to reach finality.
+// We do not want to fetch deals that are newer than the current chain head - 940 epochs.
+const finalityEpochs = 940
 const pgPool = await createPgPool()
 
 const fetchRequest = new ethers.FetchRequest(RPC_URL)
 fetchRequest.setHeader('Authorization', rpcHeaders.Authorization || '')
-const provider = new ethers.JsonRpcProvider(fetchRequest, null, { polling: true })
-
+const NAME = 'Built-in actor events'
```
juliangruber, line 22: `NAME` isn't self explanatory, what about `LOOP_NAME`?

db/lib/types.js
juliangruber: should be caught by linter or ignored

backend/lib/rpc-service/service.js
juliangruber: Should be caught by linter or ignored?

juliangruber: Although, here this style is obviously not intentional, so should be fixed independently of the linter result

backend/lib/rpc-service/service.js
```diff
@@ -0,0 +1,91 @@
...
+     * Returns actor events filtered by the given actorEventFilter
+     * @returns {Promise<Array<BlockEvent>>}
+     */
+export async function getActorEvents (actorEventFilter, makeRpcRequest) {
+  const rawEvents = (await makeRpcRequest('Filecoin.GetActorEventsRaw', [actorEventFilter]))
+  if (!rawEvents || rawEvents.length === 0) {
+    console.log(`No actor events found in the height range ${actorEventFilter.fromHeight} - ${actorEventFilter.toHeight}.`)
+    return []
+  }
+  // TODO: handle reverted events
```
juliangruber, line 39: Shall we open an issue for this TODO, and add it to the current or the next milestone?

backend/lib/rpc-service/utils.js
```diff
@@ -0,0 +1,42 @@
...
+const encodeCborInBase64 = (data) => {
+  return base64pad.baseEncode(cborEncode(data))
+}
+
+/**
+ * Converts raw event entries into a typed event
+ *
+ * @param {Array<{Key: string, Value: string}>} rawEventEntries
+ * @return {{event: Object, eventType: string}} event
+  */
```
juliangruber, line 17:
```suggestion
 */
```

juliangruber (changes requested):
Great work tackling this complexity! 👏

backend/bin/deal-observer-backend.js
```diff
@@ -1,54 +1,64 @@
...
-  observeBuiltinActorEvents
-} from '../lib/deal-observer.js'
 import { createInflux } from '../lib/telemetry.js'
+import { getChainHead, rpcRequestFn } from '../lib/rpc-service/service.js'
+import { fetchDealWithHighestActivatedEpoch, observeBuiltinActorEvents } from '../lib/deal-observer.js'
 
 const { INFLUXDB_TOKEN } = process.env
-assert(INFLUXDB_TOKEN, 'INFLUXDB_TOKEN required')
-
+// assert(INFLUXDB_TOKEN, 'INFLUXDB_TOKEN required')
```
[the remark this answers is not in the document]
NikolasHaimerl: I wanted to make it possible to run the deal observer locally too where you do not need an influx token. I'll log if the token is not in the environment.  @bajtos mentioned that the INFLUX TOKEN is stored in our github secrets, so in production the influx database should work.

backend/bin/deal-observer-backend.js
```diff
@@ -1,54 +1,64 @@
...
     try {
-      await fn()
+      // If the store is empty we set the lastEpochStore to 0 and start fetching from the current chain head
+      const currentChainHead = await getChainHead(makeRpcRequest)
+      const currentFinalizedChainHead = currentChainHead.Height - finalityEpochs
+      const lastEpochStored = (await fetchDealWithHighestActivatedEpoch(pgPool)).height ?? currentFinalizedChainHead - 1
+      if (lastEpochStored < currentFinalizedChainHead) {
+        // TODO: The free plan does not allow for fetching epochs older than 2000 blocks. We need to account for that.
+        for (const epoch of Array.from({ length: currentFinalizedChainHead - lastEpochStored + 1 }, (_, i) => i + lastEpochStored)) {
+          await observeBuiltinActorEvents(epoch, pgPool, makeRpcRequest)
```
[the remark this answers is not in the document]
NikolasHaimerl: We essentially want to keep trying with an exponential backoff and give up trying after a certain max wait time has been reached. In that case we would leave that block height alone and go to the next since we cannot retrieve anything from that block. If there are no claim events in a block then we simply not store any events for a given block height.

backend/bin/deal-observer-backend.js
```diff
@@ -1,54 +1,64 @@
...
   while (true) {
     const start = Date.now()
     try {
-      await fn()
+      // If the store is empty we set the lastEpochStore to 0 and start fetching from the current chain head
+      const currentChainHead = await getChainHead(makeRpcRequest)
+      const currentFinalizedChainHead = currentChainHead.Height - finalityEpochs
+      const lastEpochStored = (await fetchDealWithHighestActivatedEpoch(pgPool)).height ?? currentFinalizedChainHead - 1
+      if (lastEpochStored < currentFinalizedChainHead) {
+        // TODO: The free plan does not allow for fetching epochs older than 2000 blocks. We need to account for that.
```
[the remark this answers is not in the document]
NikolasHaimerl: I think that this is only relevant if we decide to also test past deals. For now we will simply use the maximum number of past blocks that we can get and build a database of deals to test. @bajtos

backend/lib/rpc-service/service.js
```diff
@@ -0,0 +1,91 @@
...
+import { ClaimEvent, RawActorEvent, BlockEvent } from './data-types.js'
+
+/** @import {CID} from 'multiformats' */
+
+/**
+ * @param {string} method
+ * @param {Object} params
+  * @returns {Promise<object>}
+  */
+export const rpcRequestFn = async (method, params) => {
```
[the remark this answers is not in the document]
NikolasHaimerl: I already use the name `makeRpcRequest` in the function names where I am parsing this function to. I renamed it to `rpcRequest`.

backend/lib/rpc-service/service.js
NikolasHaimerl: I created [one](https://github.com/filecoin-station/deal-observer/issues/22).

backend/lib/rpc-service/utils.js
```diff
@@ -0,0 +1,42 @@
...
+const rawEventEntriesToEvent = (rawEventEntries) => {
+  // Each event is defined by a list of event entries which will parsed into a typed event
+  const event = {}
+  // TODO handle if there is no type entry
```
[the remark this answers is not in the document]
NikolasHaimerl: I think there is no real way to deal with it other than marking it and exclude the event from being stored in storage. The only scenario I can think of this happening is if glif is  simply serving invalid data. The event type is part of the ActorEvent specificiation afterall. @bajtos

backend/lib/deal-observer.js
```diff
@@ -1,11 +1,86 @@
...
  * @param {Queryable} pgPool
- * @param {Provider} rpcProvider
+ * @param {Provider} makeRpcRequest
+ * @returns {Promise<void>}
+  */
+export async function observeBuiltinActorEvents (blockHeight, pgPool, makeRpcRequest) {
+  const eventType = 'claim'
+  const activeDeals = await getActorEvents(getActorEventsFilter(blockHeight, eventType), makeRpcRequest)
+  assert(activeDeals !== undefined, `No ${eventType} events found in block ${blockHeight}`)
+  console.log(`Observed ${activeDeals.length} ${eventType} events in block ${blockHeight}`)
```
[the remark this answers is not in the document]
NikolasHaimerl: The observer only fetches events per block not in block range.

backend/lib/rpc-service/service.js
```diff
@@ -0,0 +1,89 @@
...
+ */
+export const rpcRequest = async (method, params) => {
+  const reqBody = JSON.stringify({ method, params, id: 1, jsonrpc: '2.0' })
+  const response = await fetch(RPC_URL, {
+    method: 'POST',
+    headers: { 'content-type': 'application/json' },
+    body: reqBody
+  })
+
+  // @ts-ignore
```
NikolasHaimerl, line 23: I am trying to find a better solution for this right now.

backend/lib/deal-observer.js
```diff
@@ -25,18 +26,26 @@ export async function observeBuiltinActorEvents (blockHeight, pgPool, makeRpcReq
  */
 export async function fetchDealWithHighestActivatedEpoch (pgPool) {
   const query = 'SELECT * FROM active_deals ORDER BY activated_at_epoch DESC LIMIT 1'
-  const { rows } = await pgPool.query(query)
-  const result = rows.map((row) => Value.Parse(ActiveDealDbEntry, row))
+  const result = (await pgPool.query(query)).rows.map((row) => {
+    const parsedRow = Value.Parse(ActiveDealDbEntry, row)
+    parsedRow.piece_cid = fromJSON(JSON.parse(parsedRow.piece_cid))
```
bajtos, line 31: What is the format of `parsedRow.piece_cid`?

I'd like to store the CID in the database in the string form (`'baga...'`) not as DAG-JSON (`{"/": "baga..."}`).

Let's keep it simple.

backend/lib/deal-observer.js
```diff
@@ -25,18 +26,26 @@ export async function observeBuiltinActorEvents (blockHeight, pgPool, makeRpcReq
...
+ * @returns {Promise<void>}
+ * */
 export async function storeActiveDeals (activeDeals, pgPool) {
   const transformedDeals = activeDeals.map((deal) => (
     {
       activated_at_epoch: deal.height,
       miner_id: deal.event.provider,
       client_id: deal.event.client,
-      piece_cid: deal.event.pieceCid,
+      piece_cid: toJSON(deal.event.pieceCid),
```
bajtos, line 48: Same comment here, why are we converting from `CID` instance to JSON?

If it's not practical to use the `CID` type (class), then let's use the string form (`baga...`) instead.

backend/lib/rpc-service/data-types.js
```diff
@@ -4,11 +4,11 @@ const ClaimEvent = Type.Object({
   client: Type.Number(),
   provider: Type.Number(),
   pieceCid: Type.Any(),
```
bajtos, line 6: Is there a way how to use `CID` from multiformats here? For example, can we teach Typebox about `CID` as a custom type?

bajtos (changes requested):

backend/bin/deal-observer-backend.js
```diff
@@ -1,54 +1,63 @@
...
 import { createInflux } from '../lib/telemetry.js'
+import { getChainHead, rpcRequest } from '../lib/rpc-service/service.js'
+import { fetchDealWithHighestActivatedEpoch, observeBuiltinActorEvents } from '../lib/deal-observer.js'
+import assert from 'node:assert'
 
 const { INFLUXDB_TOKEN } = process.env
-assert(INFLUXDB_TOKEN, 'INFLUXDB_TOKEN required')
-
+if (!INFLUXDB_TOKEN) {
+  console.error('INFLUXDB_TOKEN was not provided. Please set the INFLUXDB_TOKEN environment variable.')
```
juliangruber, line 13: This makes it sound like the variable is required, however the program is continuing without it. What about explicitly option out of Influx behavior, by introducing a special value like on of these?

```
INFLUX_TOKEN=disable
INFLUX_TOKEN=development
INFLUX_TOKEN=dev
INFLUX_TOKEN=false
```

In this case we only need to log that `InfluxDB writing has been disabled`

backend/bin/deal-observer-backend.js
```diff
@@ -1,54 +1,63 @@
...
+}
+const LOOP_BACK_INTERVAL = 10 * 1000
```
juliangruber, line 15: What does loop back mean? It reminds me of the loopback interface.

backend/bin/deal-observer-backend.js
```diff
@@ -1,54 +1,63 @@
...
+const LOOP_NAME = 'Built-in actor events'
 const { recordTelemetry } = createInflux(INFLUXDB_TOKEN)
 
-const loop = async (name, fn, interval) => {
+const dealObserverLoop = async (makeRpcRequest, pgPool) => {
   while (true) {
     const start = Date.now()
     try {
-      await fn()
+      // If the store is empty we set the lastEpochStore to 0 and start fetching from the current chain head
```
juliangruber, line 30: I'm neither seeing a variable `lastEpochStore` (do you mean `lastEpochStored`?) nor where it would be set to 0 🤔

backend/lib/deal-observer.js
```diff
@@ -1,11 +1,102 @@
...
+  await storeActiveDeals(activeDeals, pgPool)
+}
+
+/**
+ * @param {Queryable} pgPool
+ * @returns {Promise<ActiveDealDbEntry | null>}
+ */
+export async function fetchDealWithHighestActivatedEpoch (pgPool) {
+  const query = 'SELECT * FROM active_deals ORDER BY activated_at_epoch DESC LIMIT 1'
+  const result = (await pgPool.query(query)).rows.map((row) => {
```
juliangruber, line 29: I would find this to be easier to read, as line 29 is doing a few things with ( ) as well

```suggestion
  const { rows } = await pgPool.query('SELECT * FROM active_deals ORDER BY activated_at_epoch DESC LIMIT 1')
  const result = rows.map((row) => {
```

juliangruber (changes requested):

backend/lib/deal-observer.js
NikolasHaimerl: I'll resolve this to using strings.

backend/bin/deal-observer-backend.js
NikolasHaimerl: It is using `INFLUXDB_TOKEN==='disabled'` now.

backend/bin/deal-observer-backend.js
```diff
@@ -1,54 +1,62 @@
...
+const dealObserverLoop = async (makeRpcRequest, pgPool) => {
   while (true) {
     const start = Date.now()
     try {
-      await fn()
+      const currentChainHead = await getChainHead(makeRpcRequest)
+      const currentFinalizedChainHead = currentChainHead.Height - finalityEpochs
+      // If the storage is empty we start 2000 blocks into the past as that is the furthest we can go with the public glif rpc endpoints.
+      const lastEpochStored = (await fetchDealWithHighestActivatedEpoch(pgPool)).height ?? currentChainHead.Height - 1999
+      // TODO: The free plan does not allow for fetching epochs older than 2000 blocks. We need to account for that.
```
juliangruber, line 34: Is this a merge blocker or should we open a follow up issue for this?

juliangruber (changes requested):

backend/bin/deal-observer-backend.js
NikolasHaimerl: It depends. Are we ever going to be interested in past active deals or are we fine with just using those which are discovered from when we are going to launch this service?

bajtos: I am fine with using only the deals discovered after we deploy this pull request.

Let's remove this TODO, no need to open a follow-up issue until this becomes a real-world problem.

juliangruber (approved):

backend/lib/rpc-service/utils.js
```diff
@@ -0,0 +1,49 @@
...
+      eventType = value
+      // The type entry is not part of the event itself
+      continue
+    }
+
+    // Convert CID instanes to the string representation
+    if (value[Symbol.toStringTag] === 'CID') {
+      value = value.toString()
+    } else if (typeof value !== 'number') {
+      console.error(`Unsupported type found in the raw event entries. Value enrtry: %o and key entry: ${key}, Unsupported type: ${typeof value}`, value)
```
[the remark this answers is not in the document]
bajtos: I think a message like this could be easier to work with:

```suggestion
      console.error(
        'Unsupported type %s found in the raw event entries. Key: %s Value: %o', 
        typeof value,
        key,
        value
      )
```

The key insight is that `%o` prints objects as a multi-line string, so it's best to put it at the end of the log line.

backend/package.json
bajtos: @NikolasHaimerl PTAL ☝🏻

backend/package.json
bajtos: @NikolasHaimerl please remove this line (the unused dependency).

backend/package.json
bajtos: @NikolasHaimerl please remove this line (the unused dependency).

backend/test/deal-observer.test.js
```diff
@@ -0,0 +1,78 @@
...
+      piece_cid: eventData.pieceCid,
+      piece_size: eventData.pieceSize,
+      term_start_epoch: eventData.termStart,
+      term_min: eventData.termMin,
+      term_max: eventData.termMax,
+      sector_id: eventData.sector
+    }
+    const actualData = result.rows.map((record) => {
+      return Value.Parse(ActiveDealDbEntry, record)
+    })
```
bajtos, line 55: Nitpick, feel free to ignore:

Have you considered using the `parseDeals` (loadDeals) helper here?

db/migrations/003.do.change-sector-id-type.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE active_deals
+ALTER COLUMN sector_id TYPE BIGINT;
```
bajtos, line 2: Missing EOL at EOF.

(FWIW, I believe it's possible to configure Visual Studio Code always to add EOL at EOF.)

package.json
```diff
@@ -32,5 +32,8 @@
     "prettier": "^3.4.2",
     "standard": "^17.1.2",
     "typescript": "^5.7.3"
+  },
+  "dependencies": {
+    "fastify": "^5.2.1"
```
bajtos, line 37: This does not look right to me. The production dependencies should be specified in individual sub-packages (e.g. `api/package.json`)

Please revert this block.

bajtos (changes requested):
Almost there! 💪🏻

backend/bin/deal-observer-backend.js
```diff
@@ -1,54 +1,62 @@
...
+      /* eslint-disable no-unmodified-loop-condition */
+      for (let epoch = lastEpochStored + 1; lastEpochStored <= currentFinalizedChainHead; epoch++) {
+        await observeBuiltinActorEvents(epoch, pgPool, makeRpcRequest)
+      }
```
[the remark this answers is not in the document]
juliangruber: ooh good catch!

db/migrations/003.do.change-sector-id-type.sql
juliangruber: Agreed with @NikolasHaimerl, either the Linter should catch it or we shouldn't discuss it in PRs

backend/test/deal-observer.test.js
bajtos: Interesting! IIUC, you are concerned about some external code calling `parseDeals` with a carefully crafted filter string injecting arbitrary SQL commands, right? I see how that can be a problem in a public API. 

However, the test files are entirely under our control, so I don't see any problems with exposing `parseDeals` to the tests.

What am I missing?

Also, this is a detail we can address after this pull request is landed.

db/migrations/003.do.change-sector-id-type.sql
bajtos: Fair enough, I'll stop pointing out missing EOL at EOF.

backend/bin/deal-observer-backend.js
```diff
@@ -1,54 +1,61 @@
...
+const dealObserverLoop = async (makeRpcRequest, pgPool) => {
   while (true) {
     const start = Date.now()
     try {
-      await fn()
+      const currentChainHead = await getChainHead(makeRpcRequest)
+      const currentFinalizedChainHead = currentChainHead.Height - finalityEpochs
+      // If the storage is empty we start 2000 blocks into the past as that is the furthest we can go with the public glif rpc endpoints.
+      const lastInsertedDeal = await fetchDealWithHighestActivatedEpoch(pgPool)
+      const lastEpochStored = lastInsertedDeal ? lastInsertedDeal.height : currentChainHead.Height - 1999
```
bajtos, line 34:
```suggestion
      const lastEpochStored = lastInsertedDeal ? lastInsertedDeal.height : currentChainHead.Height - maxPastEpochs
```

backend/bin/deal-observer-backend.js
```diff
@@ -1,54 +1,61 @@
...
+    console.log(`Loop "${LOOP_NAME}" took ${dt}ms`)
 
-    recordTelemetry(`loop_${slug(name, '_')}`, point => {
-      point.intField('interval_ms', interval)
-      point.intField('duration_ms', dt)
-    })
-
-    if (dt < interval) {
-      await timers.setTimeout(interval - dt)
+    if (!INFLUXDB_TOKEN) {
```
bajtos, line 46: Sorry, I made a mistake in my previous suggestion. We want to record the telemetry when the token is set. (But print a warning when it is _not_ set.)

```suggestion
    if (INFLUXDB_TOKEN) {
```

backend/lib/rpc-service/service.js
bajtos: Nitpick:

We usually mention the issue in the code comment.

```suggestion
  // TODO: handle reverted events
  // https://github.com/filecoin-station/deal-observer/issues/22
```

backend/test/deal-observer.test.js
NikolasHaimerl: Yes, I would try to avoid exporting functions which we do not have to export especially if these functions allow for modifications in our database.

backend/test/deal-observer.test.js
```diff
@@ -0,0 +1,78 @@
...
+      pieceSize: 4n,
+      termStart: 5,
+      termMin: 12340,
+      termMax: 12340,
+      sector: 6n,
+      payload_cid: null // not present in event data
+    }
+    const event = Value.Parse(BlockEvent, { height: 1, event: eventData, emitter: 'f06' })
+
+    // @ts-ignore
```
bajtos, line 39: I created a PR against your feature branch to fix the types so that you don't need to disable type-checking.

PTAL: https://github.com/filecoin-station/deal-observer/pull/27


@NikolasHaimerl You can merge that pull request to bring my changes to your branch & pull request.

bajtos (changes requested):
I reviewed the latest version and have a few more comments to address.

Also not all of my previous comments have been addressed, PTAL. 

_Remember to expand the hidden conversations as shown on this screenshot:_

Screenshot 2025-01-27 at 10 57 52

db/migrations/003.do.change-sector-id-type.sql
bajtos: (I also added a note to https://github.com/space-meridian/roadmap/issues/166)

backend/package.json
```diff
@@ -14,11 +14,14 @@
...
     "@sentry/node": "^8.48.0",
+    "@sinclair/typebox": "^0.34.14",
     "@types/slug": "^5.0.9",
     "debug": "^4.4.0",
-    "ethers": "^6.13.5",
-    "pg": "^8.13.1",
-    "slug": "^10.0.0"
```
bajtos, line 22: If you are not using `slug` anymore, then please remove `@types/slug` above as well.

backend/lib/deal-observer.js
```diff
@@ -1,11 +1,110 @@
...
+      transformedDeals.map(deal => deal.term_min),
+      transformedDeals.map(deal => deal.term_max),
+      transformedDeals.map(deal => deal.sector_id)
+    ])
+
+    // Commit the transaction if all inserts are successful
+    console.log(`Inserting ${activeDeals.length} deals took ${Date.now() - startInserting}ms`)
+  } catch (error) {
+    // If any error occurs, roll back the transaction
+    console.error('Error inserting deals:', error.message)
```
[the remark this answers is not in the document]
NikolasHaimerl: I added an issue for this: https://github.com/filecoin-station/deal-observer/issues/28

backend/package.json
bajtos: FWIW, I created a tech-debt task to setup a linter rule to flag unused dependencies: https://github.com/space-meridian/roadmap/issues/221

backend/test/deal-observer.test.js
NikolasHaimerl: Thanks I merged it.

backend/bin/deal-observer-backend.js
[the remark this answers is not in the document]
bajtos: @NikolasHaimerl It's not clear to me how you resolved this comment. Could you please explain?

bajtos (approved):
Please make sure the loop iterating over multiple epochs is written correctly, see https://github.com/filecoin-station/deal-observer/pull/12/files#r1928634351

Other than that, the pull request looks great now! 👌🏻 

No further review is necessary as far as I am concerned, merge at will.

NikolasHaimerl:
@bajtos 

> Please make sure the loop iterating over multiple epochs is written correctly, see https://github.com/filecoin-station/deal-observer/pull/12/files#r1928634351
> 
> Other than that, the pull request looks great now! 👌🏻
> 
> No further review is necessary as far as I am concerned, merge at will.

I believe I committed your code suggestion. Which is why the comment was resolved. But I see that the loop is incorrect. I"ll fix that.

PR #13, opened 2025-01-17
feat: poc retrieve ddo claims in loop

NikolasHaimerl:

1. Implement a POC of the backend for the deal observer. The binary retrieves and stores DDO deals in the database in a loop.


backend/bin/deal-observer-backend.js
```diff
@@ -7,37 +7,53 @@ import slug from 'slug'
...
 
-const { INFLUXDB_TOKEN } = process.env
-assert(INFLUXDB_TOKEN, 'INFLUXDB_TOKEN required')
+//const { INFLUXDB_TOKEN } = process.env
+//assert(INFLUXDB_TOKEN, 'INFLUXDB_TOKEN required')
 
 const pgPool = await createPgPool()
+// Filecoin will need some epochs to reach finality. 
+// We do not want to fetch deals that are newer than the current chain head - 900 epochs.
+const finalityEpochs = 900
```
bajtos, line 20: It's great you thought about this!

I double-checked notes from my the conversation with Rod; he recommended waiting for 940 blocks (~8 hours).

backend/lib/deal-observer.js
```diff
@@ -1,11 +1,136 @@
...
+          client,
+          piece_cid,
+          piece_size,
+          term_start_epoch,
+          term_min,
+          term_max,
+          sector,
+          payload_cid
+        )
+        VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9 ,$10)
```
bajtos, line 79: Running one SQL query per deal will take a long time to complete when there are hundreds to thousands of deals per block.

We should insert deals in one SQL query, check how spark-evaluate does it with stats.

backend/lib/deal-observer.js
```diff
@@ -1,11 +1,136 @@
...
+    }
+  }
+  async fetchActiveDealsByKey(searchParams){
+    function buildQuery(baseQuery, conditions) {
+      const queryParts = [];
+      
+      // Iterate through conditions and build the WHERE clauses dynamically
+      conditions.forEach(({key,value}) => {        
+        // Add the condition to the queryParts array
+        queryParts.push(`${key} = ${value}`);
```
bajtos, line 118: This is an SQL-injection vulnerability.

AFAICT, `fetchActiveDealsByKey` is not used anywhere, can we remove it?

backend/bin/deal-observer-backend.js
NikolasHaimerl: Updated!

backend/lib/deal-observer.js
NikolasHaimerl: I created caching. We can adjust the number of deals to be cached. Usually it is between 20-70 claim events per block right now.

backend/lib/deal-observer.js
NikolasHaimerl: Removed!

PR #16, opened 2025-01-19
Add beryx to JSON-RPC providers

juliangruber:
https://beryx.io/rpc

So far, this provider seems to work well for us.


bajtos (approved):
:shipit: 

LGTM, as long as we are not using FallbackProvider, but rotating the provider URL to use. (I believe that's the case in this repo.)

PR #17, opened 2025-01-21
api: switch to `fastify`

juliangruber:


api/bin/deal-observer-api.js
```diff
@@ -11,15 +9,10 @@ const {
...
-const server = http.createServer(handler)
+const app = createApp({
+  pgPool,
+  logger: ['1', 'true'].includes(REQUEST_LOGGING)
+})
 console.log('Starting the http server on host %j port %s', HOST, PORT)
-server.listen(Number(PORT), HOST)
-await once(server, 'listening')
-console.log(`http://${HOST}:${PORT}`)
+console.log(await app.listen({ host: HOST, port: Number(PORT) }))
```
bajtos, line 18: This line does two things - it starts the server and logs some info, presumably about the base URL. I prefer to split it into two lines, if you don't mind.

```suggestion
const serverUrl = await app.listen({ host: HOST, port: Number(PORT) })
console.log(serverUrl)
```

api/test/handler.test.js
```diff
@@ -1,44 +1,31 @@
...
 
-    const handler = createHandler({
+    app = createApp({
       pgPool,
-      logger: {
-        info: debug,
-        error: console.error,
-        request: debug
-      }
+      logger: false
```
bajtos, line 21: This is a "breaking" change. Before, tests printed info messages via the `debug` channel, thus keeping the test output clean. After your change, all info messages will be printed to the console, potentially polluting the test output.

Can we preserve the old behaviour please?

bajtos (changes requested):
NOICE!

api/test/handler.test.js
juliangruber: I was concerned about this as well, but didn't look into it further out of laziness. I will fix it :)

juliangruber: https://github.com/filecoin-station/deal-observer/pull/17/commits/36ef539a3123011c583f1574cceb323fbed95dd0

bajtos (approved):
👏🏻

PR #29, opened 2025-01-27
chore: modify logging, increase loop timout, add exponential backoff

NikolasHaimerl:
This PR adds the following changes:

1. Add an exponential backoff method for the fetching of apis
2. Add logging to API requests
3. Remove logging of deals when no deals were discovered
4. Remove assertion during deal fetching 
5. Increase timeout between loop iterations
6. Comment out zondax API until we have the token configured


juliangruber:
> This PR adds the following changes:
>
> Add an exponential backoff method for the fetching of apis
Add logging to API requests
Remove logging of deals when no deals were discovered
Remove assertion during deal fetching
Increase timeout between loop iterations
Comment out zondax API until we have the token configured

Could you please rename the PR to signify that this is about more than logging? Otherwise, if we squash and merge this as is, we will have an entry in our git history that reads "chore: add logging for deal observer" but does so much more - which can be confusing

backend/bin/deal-observer-backend.js
```diff
@@ -12,7 +12,7 @@ const { INFLUXDB_TOKEN } = process.env
 if (!INFLUXDB_TOKEN) {
   console.error('INFLUXDB_TOKEN not provided. Telemetry will not be recorded.')
 }
-const LOOP_INTERVAL = 10 * 1000
+const LOOP_INTERVAL = 100 * 1000
```
juliangruber, line 15: what is the motivation for this change? Side note: There is an easier way to write this:

```suggestion
const LOOP_INTERVAL = 100_000
```

backend/bin/deal-observer-backend.js
```diff
@@ -32,6 +32,7 @@ const dealObserverLoop = async (makeRpcRequest, pgPool) => {
       // If the storage is empty we start 2000 blocks into the past as that is the furthest we can go with the public glif rpc endpoints.
       const lastInsertedDeal = await fetchDealWithHighestActivatedEpoch(pgPool)
       const lastEpochStored = lastInsertedDeal ? lastInsertedDeal.height : currentChainHead.Height - maxPastEpochs
+      console.log(`Last epoch stored: ${lastEpochStored}, current chain head: ${currentChainHead.Height}, current finalized chain head: ${currentFinalizedChainHead}`)
```
juliangruber, line 35: What is the motivation for adding this log line? Do we always need this, or should we use https://github.com/debug-js/debug instead, so we can enable and disable in prod on demand?

backend/lib/config.js
```diff
@@ -1,6 +1,9 @@
 const {
-  RPC_URLS = 'https://api.node.glif.io/rpc/v0,https://api.zondax.ch/fil/node/mainnet/rpc/v1',
-  GLIF_TOKEN
+  RPC_URLS = 'https://api.node.glif.io/rpc/v0',
+  // TODO: Enable API once token is added to the environment
```
juliangruber, line 3: Which token?

juliangruber: Which token? I'm not aware of Zondax RPC endpoints requiring auth 🤔

backend/lib/deal-observer.js
```diff
@@ -16,8 +15,9 @@ import { Value } from '@sinclair/typebox/value'
 export async function observeBuiltinActorEvents (blockHeight, pgPool, makeRpcRequest) {
   const eventType = 'claim'
   const activeDeals = await getActorEvents(getActorEventsFilter(blockHeight, eventType), makeRpcRequest)
-  assert(activeDeals !== undefined, `No ${eventType} events found in block ${blockHeight}`)
-  console.log(`Observed ${activeDeals.length} ${eventType} events in block ${blockHeight}`)
+  if (activeDeals.length > 0) {
+    console.log(`Observed ${activeDeals.length} ${eventType} events in block ${blockHeight}`)
+  }
```
juliangruber, line 20: Is it still useful to call `storeActiveDeals` with no deals?

backend/lib/rpc-service/service.js
```diff
@@ -14,13 +15,18 @@ import { ClaimEvent, RawActorEvent, BlockEvent, RpcRespone } from './data-types.
...
   const reqBody = JSON.stringify({ method, params, id: 1, jsonrpc: '2.0' })
-  const response = await fetch(RPC_URL, {
-    method: 'POST',
-    headers: { 'content-type': 'application/json' },
-    body: reqBody
-  })
-
-  return Value.Parse(RpcRespone, (await response.json())).result
+  try {
+    const response = await fetchWithExponentialBackoff(async () => {
```
juliangruber, line 19: This can be implemented without bespoke functions, using https://github.com/sindresorhus/p-retry. Here is how we've been using it so far: https://github.com/search?q=org%3Afilecoin-station%20p-retry&type=code

juliangruber (changes requested):

backend/bin/deal-observer-backend.js
NikolasHaimerl: The motivation was that during running the application we saw a lot of prints for how long the loop lasted. There were two options: 

1. Increase the loop duration
2. Only print the loop duration very X rounds

The latter would result in a loss of information which is why I chose the former. Inside the loop we retrieve all blocks up to the current head anyway, so it does not matter that much whether we ask for blocks sooner or later. 
The Filcoin block time is 30 seconds per block and the loop fetches up to all blocks between the current database and the head (minus finality).

backend/bin/deal-observer-backend.js
NikolasHaimerl: I don't have a strong opinion on which library we use for logging. If we usually go with debug-js then I am fine with that. 
The line was added to account for deletion of others and summarizing the key stats of each loop round in one output.

backend/lib/config.js
NikolasHaimerl: We are currenty being rate limited by the Zondax RPC API. For GLIF there exist tokens which solve the rate limiting to a large extend. I have already added that one to the Fly environment. I am unsure about the Zondax RPC API though. I only saw API keys for their data API, not their node API. If we cannot get a key we should think about removing zondax from the list of RPC nodes.

backend/bin/deal-observer-backend.js
juliangruber: There's a difference in these libraries. `debug` only prints its lines when you tell it to, via environment variables. Usually, these `debug()` calls will be no ops - only when we pass `DEBUG=deal-observer` or similar will it print out these lines.

Do you think this is a line that should always be logged, or only on demand?

backend/lib/rpc-service/service.js
NikolasHaimerl: Thanks for the hint, I'll use it for this PR as well.

backend/lib/deal-observer.js
NikolasHaimerl: Good point, I don't think it is.

backend/lib/config.js
juliangruber: I don't think they offer a way to get more requests (yet). Do you have an example of rate limiting from them (eg: url, call frequency, error code)? Then I'll reach out to them via email.

backend/bin/deal-observer-backend.js
NikolasHaimerl: I see, this would be a line we probably want on demand to track the progress of the loop.

backend/lib/config.js
NikolasHaimerl: I do not. I only saw that the returned data started to be `undefined` after a thousand calls or so. This also only occurred during production as in development I used GLIF with an API key. After deploying this change we should be able to see the returned error message though.

backend/lib/deal-observer.js
```diff
@@ -16,9 +15,10 @@ import { Value } from '@sinclair/typebox/value'
 export async function observeBuiltinActorEvents (blockHeight, pgPool, makeRpcRequest) {
   const eventType = 'claim'
   const activeDeals = await getActorEvents(getActorEventsFilter(blockHeight, eventType), makeRpcRequest)
-  assert(activeDeals !== undefined, `No ${eventType} events found in block ${blockHeight}`)
-  console.log(`Observed ${activeDeals.length} ${eventType} events in block ${blockHeight}`)
-  await storeActiveDeals(activeDeals, pgPool)
+  if (activeDeals.length > 0) {
+    console.log(`Observed ${activeDeals.length} ${eventType} events in block ${blockHeight}`)
+    await storeActiveDeals(activeDeals, pgPool)
+  }
```
bajtos, line 21: I find it important for the logs to tell us when we successfully fetched & processed actor events, but there were no deals found, as a confirmation that the process finished.

I also think the short-circuit handling of empty list of active deals is better handled inside `storeActiveDeals` function. 1) Every place calling `storeActiveDeals()` benefit from it. 2) It's easier to write a test to verify this behaviour.

backend/lib/rpc-service/service.js
```diff
@@ -14,13 +15,18 @@ import { ClaimEvent, RawActorEvent, BlockEvent, RpcRespone } from './data-types.
...
-  return Value.Parse(RpcRespone, (await response.json())).result
+  try {
+    const response = await pRetry(async () => {
+      return await fetch(RPC_URL, {
+        method: 'POST',
+        headers: { 'content-type': 'application/json' },
+        body: reqBody
+      })
+    }, { retries: 5 })
+    return Value.Parse(RpcRespone, await response.json()).result
```
bajtos, line 26: You should check that the response status is 200. AFAIK `fetch()` does not throw an error in such a case.

We usually use a helper like `assertOkResponse()` that includes the response body in the error message, see https://github.com/filecoin-station/piece-indexer/blob/main/indexer/lib/http-assertions.js

```js
const response = await fetch(/*...*/)
await assertOkResponse(response)
const body = await response.json()
// etc.
```

You should also consider which error status codes should trigger a retry. If we retry a rate-limited request, will the situation worsen because we are increasing the rate of requests even more?

bajtos (changes requested):
I have few comments & suggestions to discuss, see below.

backend/bin/deal-observer-backend.js
```diff
@@ -12,7 +12,7 @@ const { INFLUXDB_TOKEN } = process.env
 if (!INFLUXDB_TOKEN) {
   console.error('INFLUXDB_TOKEN not provided. Telemetry will not be recorded.')
 }
-const LOOP_INTERVAL = 10 * 1000
+const LOOP_INTERVAL = 100_000
```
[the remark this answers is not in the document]
NikolasHaimerl: No particular reason other than choosing an interval which will not result in too many iterations per minute. 
We can use the 30 seconds, although I don't think it makes much of a difference since the loop fetches multiple blocks at a time if the current local chain head is behind the filecoin chain head.

backend/bin/deal-observer-backend.js
NikolasHaimerl: Wouldn't we have to rerun a binary while setting the DEBUG variable? 
I am not sure whether one can toggle between showing debug output and not showing it while an application is running.

NikolasHaimerl: Is there a significant gain by using `debug` over `console.debug(..)` and simply filtering them out in grafana?

backend/lib/rpc-service/service.js
NikolasHaimerl: Thanks I'll reuse that helper method.

backend/bin/deal-observer-backend.js
juliangruber: If we log everything it will increase our Papertrail bill, and make logs harder to read.

I think we could switch to a proper logging library eventually (as @bajtos also suggested), which uses syslog levels, and then we could re-evaluate including debug level log lines. For now, this is the system we have, and changing it is out of scope of this PR (you want discussion to stay focused).

If you would like to suggest a change to our logging setup, a great place is either a new GitHub issue in https://github.com/space-meridian/roadmap, or a new thread in Slack

backend/lib/config.js
juliangruber: > I do not

What does this refer to?

>  I only saw that the returned data started to be undefined after a thousand calls or so

Do you know if the HTTP status also changed? If they return status 200 but a bad response, that would be alarming.

> This also only occurred during production as in development I used GLIF with an API key

Got it, so merging this PR will give us more insight into rate limiting?

backend/bin/deal-observer-backend.js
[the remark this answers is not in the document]
juliangruber: > other than choosing an interval which will not result in too many iterations per minute.

How is "too many iterations per minute" defined? Can we formalize this?

backend/bin/deal-observer-backend.js
bajtos: > this would be a line we probably want on demand to track the progress of the loop.

I propose using the `debug` module for logging this line and moving on to finish this PR soon 🙏🏻

backend/lib/config.js
bajtos: > I only saw that the returned data started to be `undefined` after a thousand calls or so. This also only occurred during production as in development I used GLIF with an API key. After deploying this change we should be able to see the returned error message though.

From my perspective, it's crucial to understand what exactly happened after those ~thousands calls - what response the server returned? It could have been 429 Too Many Requests but it could have been something else as well. We don't know that now, so the first thing should be improving our logging (or observability in general) so that the next time we encounter this problem, we have more information about the root cause.

I guess this pull request addresses my request since you are adding a console.error log to `rpcRequest` 👍🏻 

The second issue I see is that our code did not detect this error; it treated "RPC API error response" the same way as "no events found in the block".  This is a red flag to me, because it conceals a problem that affects the quality of our data.

For example, in the logs from the initial run, I noticed that every N blocks, where N is somewhere between 10-100 (I don't remember the exact value), we would find no `claim` events in that block. That's suspicious on its own - why would a tiny fraction of blocks don't activate any FIL+ sector, when almost all blocks activate ~10-30 of them? Now that it seems we were treating RPC API errors the same way as "no events found", I don't have enough confidence that we correctly processed the events on the first run.

Let's make sure we have a good understanding of how our implementation works, backed by tests and/or production logs.

backend/lib/rpc-service/service.js
```diff
@@ -14,13 +15,18 @@ import { ClaimEvent, RawActorEvent, BlockEvent, RpcRespone } from './data-types.
...
+  } catch (e) {
+    console.error(`Failed to make RPC request: ${e.message}`)
+  }
```
bajtos, line 29: IIUC, when there is an error, this code prints the error to stderr and the function `rpcRequest` returns `undefined`. This means all callers of `rpcRequest` must understand this path and handle the `undefined` result.

I guess this works well in Rust, where you have `Result<T, E>` and the compiler tells you when you forget to handle the error path.

We don't have such guarantees in JavaScript. TypeScript _may_ be able to detect places where we don't check for `undefined`, but it's not reliable enough (we have `any` type in too many places), it's not enforced at runtime, and it's also difficult to spot during code reviews.

What we usually do: When we don't have a specific way how to handle the error, we either don't catch it or re-throw it with more details, letting the caller (or the caller of the caller) to deal with it.

bajtos: So in this particular case, I suggest the following code here, plus make sure we log unhandled errors in the top-level loop.

```suggestion
  } catch (e) {
    e.message = `Failed to make RPC request for ${method}: ${e.message}`
    throw e
  }
```

Also note that JavaScript errors include a stack trace that is often very useful for troubleshooting the issue, so it's best to log the entire error object (and let the console handle formatting for display), not only the error message. You cannot use a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) for that, you have to pass values to format as additional arguments to console methods.

```js
console.error('Failed to make RPC request:', e)
```

backend/lib/deal-observer.js
bajtos: Here is what I would do myself:

```diff
 export async function storeActiveDeals (activeDeals, pgPool) {
+  if (!activeDeals.length) return

   const transformedDeals = activeDeals.map((deal) => (
```

backend/lib/rpc-service/service.js
bajtos: I see you already have code in the top-level loop to report errors, all is good there 👍🏻 

https://github.com/filecoin-station/deal-observer/blob/492ff6aabbf10e38ee53a9c33e17bdeac83c1f06/backend/bin/deal-observer-backend.js#L38-L41

backend/bin/deal-observer-backend.js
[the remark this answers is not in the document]
bajtos: I agree that from the functional perspective, it's less important how often the loop reiterates.

I am more concerned about the feedback loop length and the developer experience.

If we process new events every 100 seconds, and I am watching production logs to troubleshoot a problem, having to wait for 100 seconds to see the outcome of the next iteration is a bit annoying. It's nothing major, but it's still a tiny papercut. That's why I prefer shorter delays, e.g. 30 seconds matching the filecoin block time.

Having written that, this is a detail that's easy to change later; I am fine with using the proposed delay of 100 seconds.

Issue #30, opened 2025-01-27
Piece Indexer Missing CIDs


NikolasHaimerl:
Many miners don't advertise their deals to IPNI at all. This leads to the piece CID indexer not keeping record of payloads for all piece CID and minerID combinations. 

It is possible that miners do publish to IPNI but it takes an unexpectedly long amount of time. 
To summarize three cases need to be accounted for:

1. Miner does not publish to IPNI -> Payload is not retrievable
2. Miner does publish to IPNI but it is not yet updated -> Wait for update
3. Miner does publish to IPNI and it has been updated -> Fetch payload.

NikolasHaimerl:
> > Many miners don't advertise their deals to IPNI at all. This leads to the piece CID indexer not keeping record of payloads for all piece CID and minerID combinations.
> 
> Could you please explain why this is an issue?
> 
> In my mind, when deal-observer asks PieceIndexer for a payload CID and receives back "not found" response, it should retry the request after some time to take into account IPNI ingestion delays, and eventually give up. E.g. retry in 2 hours, 8 hours, 24 hours and then give up.
> 
> This seems like a simpler solution to me:
> 
>     * We already need to implement "wait for update" inside deal-observer, so my suggestion does not make deal-observer any more complex.
> 
>     * We don't need to query the list of providers - less code to write & maintain.

It is an issue in sofar, as the current implementation expects all piece CIDs to be present in the IPNI piece indexer. 
I had a similar solution in mind. One thing that we need to think of is whether we want deals to where there is simply no Payload published on IPNI to be retried for at a later point in time, or whether we should flag them. That latter would guarantee that when we restart the piece indexer part of the deal-observer it does not try to fetch deals it had already given up on before.

bajtos:
>One thing that we need to think of is whether we want deals to where there is simply no Payload published on IPNI to be retried for at a later point in time, or whether we should flag them. That latter would guarantee that when we restart the piece indexer part of the deal-observer it does not try to fetch deals it had already given up on before.

Yes, we need to flag deals where we already contacted PieceIndexer/IPNI and could not find any payload CID.

The current DB of all active f05 deals has over 20 million entries. From what I see in Spark measurements, around 65% of retrievals fail because the miner did not advertise the deal to IPNI. That gives us around 13 million deals not advertised to IPNI. We don't want to retry 13 million IPNI/PieceIndexer queries.

> when we restart the piece indexer part of the deal-observer

[...]

PR #31, opened 2025-01-27
feat: fetch payload CIDs from piece indexer

NikolasHaimerl:
This MR proposes the following changes:

1. Add a piece cid indexer which fetches miner IDs and payloads for miner IDs and piece CIDs
2. Add tests for the piece indexer loop
3. Refactor database access for better code visibility

The lookup operation for piece CIDs is not yet complete. The piece cid indexer does not have all combinations of minerID and piece CID for any built-in actor event. A follow-up PR which deals with this issue is described in this [issue](https://github.com/filecoin-station/deal-observer/issues/30). 
We do not want to add the piece indexer loop to the list of executables for now. This should be done as soon as missing payloads are also dealt with.


backend/bin/deal-observer-backend.js
```diff
@@ -1,60 +1,30 @@
...
-        point.intField('duration_ms', dt)
-      })
-    }
-    if (dt < LOOP_INTERVAL) {
-      await timers.setTimeout(LOOP_INTERVAL - dt)
-    }
-  }
-}
-
 await dealObserverLoop(
```
NikolasHaimerl, line 19: Extracting the continous loops adds more clarity to the overall code structure and allows for better testing of the loop.

backend/bin/piece-indexer-backend.js
```diff
@@ -0,0 +1,25 @@
...
+await pieceIndexerLoop(
+  rpcRequest,
+  pixRequest,
+  await createPgPool(),
+  {
+    loopInterval: 10 * 1000,
+    recordTelemetry,
+    loopName: 'Piece Indexer',
+    influxToken: INFLUXDB_TOKEN,
+    queryLimit: 100000
```
NikolasHaimerl, line 23: A limit on how many database entries we want to fetch with one query call. We use this limit so memory does not get full when dealing with lots of entries that are being querried.

backend/test/deal-observer.test.js
```diff
@@ -14,6 +18,7 @@ describe('deal-observer-backend', () => {
   })
 
   after(async () => {
+    await pgPool.query('DELETE FROM active_deals')
```
NikolasHaimerl, line 21: Make sure that no data is left in the database after the tests have run.

backend/bin/deal-observer-backend.js
```diff
@@ -1,61 +1,30 @@
 import { createPgPool } from '@filecoin-station/deal-observer-db'
-import * as Sentry from '@sentry/node'
```
juliangruber, line 2: Sentry should stay in the `bin/*` files, so that we know the first thing that every service entry point does is initialize error reporting

backend/bin/loops.js
juliangruber, line 1: It would have been nice to perform this refactor in a separate PR, so that it's easier to review which changes are logic changes and which are code structure changes

juliangruber: Could you please highlight the business logic changes in this PR, using GitHub comments?

NikolasHaimerl: I can still do this, this is a Draft after all. I would have appreciated some early feedback since we are under a time constraint. 
I'll open separate PRs for the refactorings and merge them before I merge this PR. WDYT?

juliangruber: I'm very happy to give happy feedback! What should I give early feedback on? The whole diff?

backend/bin/piece-indexer-backend.js
```diff
@@ -0,0 +1,25 @@
...
+if (!INFLUXDB_TOKEN) {
+  console.error('INFLUXDB_TOKEN was not provided. Please set the INFLUXDB_TOKEN environment variable.')
+}
+const { recordTelemetry } = createInflux(INFLUXDB_TOKEN)
+
+await pieceIndexerLoop(
+  rpcRequest,
+  pixRequest,
+  await createPgPool(),
+  {
```
bajtos, line 18: We usually don't mix positional and named parameters. Let's move all parameters to the param object.

```js
await pieceIndexerLoop({
  rpcRequest,
  pixRequest,
  pgPool: await createPgPool(),
  loopInterval: 10 * 1000,
  // etc.
})

File: backend/lib/config.js [hunk 6]
```diff
@@ -1,9 +1,7 @@
 const {
-  RPC_URLS = 'https://api.node.glif.io/rpc/v0',
-  // TODO: Enable API once token is added to the environment
-  // https://api.zondax.ch/fil/node/mainnet/rpc/v1',
-  GLIF_TOKEN,
-  ZONDAX_TOKEN
+  RPC_URLS = 'https://api.node.glif.io/rpc/v0,https://api.zondax.ch/fil/node/mainnet/rpc/v1',
+  PIECE_INDEXER_URL = 'https://pix.filspark.com',
```
bajtos, line 3: When calling this URL, the requests will go from Fly to Cloudflare back to Fly, which may increase our ingress/egress bills.

Do you have any concerns about that, @juliangruber?

A more performant option is to route the requests internally within Fly.io only, see e.g. https://fly.io/docs/networking/app-services/#fly-io-internal-dns-and-6pn-addresses

That would require us to change our PIX server to allow requests for two hostnames - the public one at https://pix.filspark.com and the Fly-private one. See https://github.com/filecoin-station/piece-indexer/blob/1b97e24b1a972401992a57d6bc7c23f9b17e17d7/api/lib/app.js#L15-L19

Considering the complexity and time pressure, I think it's best to use https://pix.filspark.com for now.

backend/bin/piece-indexer-backend.js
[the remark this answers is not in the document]
NikolasHaimerl: Based on our conversation at the Colab last week I was under the impression we want two separately running loops one fetching the new active deals and the other that updates the payloads (technically we'll also have a third one with the spark-api updates). 
I am happy to include them in the same process as described by you. 
I too think this would mitigate the complexity of handling two processes. 
let me make the change then.

backend/bin/loops.js
```diff
@@ -0,0 +1,82 @@
...
+        }
+      } else {
+        const dealsWithMissingPayloadCid = await fetchDealsWithNoPayloadCid(pgPool, processingBlockHeight, options.queryLimit)
+        if (dealsWithMissingPayloadCid === null) {
+          console.log('No deals with missing payload CID found')
+        } else {
+          console.log(`Found ${dealsWithMissingPayloadCid.length} deals with missing payload CID`)
+          await updatePayloadCids(pgPool, rpcRequest, dealsWithMissingPayloadCid, pixRequest)
+          processingBlockHeight = dealsWithMissingPayloadCid[dealsWithMissingPayloadCid.length - 1].activated_at_epoch
+        }
```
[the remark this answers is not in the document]
NikolasHaimerl: There is always the possibility that some deals just don't have a payload. If we simply iterated over all deals that do not have payloads we will stumble over the same deals again and again. Furthermore, we do not want to retrieve all deals with missing payloads all at once, as that would be dangerous for our memory consumption. So keeping track of which deals we have looked at and which we have not is important.

backend/test/deal-observer.test.js
```diff
@@ -32,11 +37,11 @@ describe('deal-observer-backend', () => {
       termMin: 12340,
       termMax: 12340,
       sector: 6n,
-      payload_cid: null // not present in event data
+      payload_cid: undefined // not present in event data
```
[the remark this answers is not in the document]
bajtos: +1 to use `undefined` on the JavaScript side.

Please double-check that when we fetch data from PostgreSQL, the `NULL` values are converted to `undefined` by [node-postgres](https://node-postgres.com/), our database client.

db/lib/database-access.js
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/**
+   * @param {Queryable} pgPool
+   * @param {number} fromBlockHeight
+   * @param {number} limit
+   * @returns {Promise<Array<ActiveDealDbEntry>>}
+   */
+export async function fetchDealsWithNoPayloadCid (pgPool, fromBlockHeight, limit) {
+  const query = `SELECT * FROM active_deals WHERE payload_cid IS NULL AND activated_at_epoch >= ${fromBlockHeight} ORDER BY activated_at_epoch ASC LIMIT ${limit}`
```
bajtos, line 123: I am not comfortable injecting raw string values into SQL queries, because it creates SQL injection vulnerability.

I can see how this may be safe to do within the current pull request, but I am concerned that future changes may not be aware that `fromBlockHeight` and `limit` must be sanitised for SQL.

Proposal: add runtime assertions to verify that `fromBlockHeight` and `limit` are numbers.

```suggestion
  assert(typeof fromBlockHeight === 'number', 'fromBlockHeight must be a number')
  assert(typeof limit === 'number', 'limit must be a number')
  const query = `SELECT * FROM active_deals WHERE payload_cid IS NULL AND activated_at_epoch >= ${fromBlockHeight} ORDER BY activated_at_epoch ASC LIMIT ${limit}`
```

Our preferred solution is to use parameterised SQL queries. 

Unfortunately, the `loadDeals` helper does not support that. To me, it's a sign that `loadDeals` is not the right abstraction to use.

Ideally, we should be running code like this (possible extracting some parts into helper functions):

```js
const { rows } = await pgPool.query(`
  SELECT * FROM active_deals
  WHERE 
    payload_cid IS NULL 
    AND activated_at_epoch >= $1
  ORDER BY activated_at_epoch ASC 
  LIMIT $2
  `, [
    fromBlockHeight,
    limit
  ])

const result = rows.map(deal => {
  // SQL used null, typebox needs undefined for null values
  Object.keys(deal).forEach(key => {
    if (deal[key] === null) {
      deal[key] = undefined
    }
  })
  return Value.Parse(ActiveDealDbEntry, deal)
})
 
return result.length > 0 ? result : undefined
```

Looking at the code above, I would say the right abstraction is to create a function accepting `rows` returned by `pgPool.query` and parsing them to an array of ActiveDealDbEntry objects.

backend/bin/loops.js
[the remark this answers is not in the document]
bajtos: After reading most of the pull request and your comment in https://github.com/CheckerNetwork/deal-observer/issues/30#issuecomment-2620992594, I think I am starting to understand your design.

Here is my understanding:

- You use the combination of `payload_cid` and `activated_at_epoch` to decide which deals should be updated.
- You process the deals from the oldest epoch in small batches and progress towards  more recent deals.
- After processing a batch of deals, you look at the latest `activated_at_epoch` value in the processes deals, and fetch deals created on or after that epoch in the next iteration.

I see a major problem with that:
- When we restart the loop, e.g when restarting the process after a new version was deployed, we will start from the oldest epoch in our database and re-try the PieceIndexer query for every deal for which we don't have the payload CID yet.

As I explained in https://github.com/CheckerNetwork/deal-observer/issues/30#issuecomment-2621576648:
- Around ~63% deals are not indexed. 
- The current DB of all active f05 deals has over 20 million entries. 
- That gives us ~13 million deals for which we cannot obtain the piece CID.

Of course, deal-observer does not have so many deals in the database yet. But the number is growing quickly.
- fil-deal-ingester is adding >100k new days daily.
- Every f05 deal emits the `claim` event, so it will be tracked by deal-observer too.
- That means deal-observer will every day add >60k deals for which we will retry the PieceIndexer query on service restart.

> If we simply iterated over all deals that do not have payloads we will stumble over the same deals again and again. 

Agreed. As I understand your code, this will still be the case after service restart.

> Furthermore, we do not want to retrieve all deals with missing payloads all at once, as that would be dangerous for our memory consumption. So keeping track of which deals we have looked at and which we have not is important.

I agree we must rate-limit the PieceIndexer queries. Besides the memory usage, we also need to consider the load we put on the PieceIndexer service and IPNI in the future.

---

A possible solution if you want to keep the design based on epoch:

When restarting the service, find the first epoch where all deals are missing payload CID, and start from that.

I think that's good enough for the initial version.

backend/lib/config.js
juliangruber: I agree with using Cloudflare, if we add an issue for cleaning this up later

backend/lib/pix-service/service.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+* @param {number} providerId
+* @param {string} pieceCid
+* @returns {Promise<string>}
+*/
+export async function fetchPayloadCid (providerId, pieceCid, rpcRequest, pixRequest) {
+  let minerPeerId
+  // Check for cached miner peer IDs
+  if (minerPeerIdsCache.has(providerId)) {
```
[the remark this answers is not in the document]
juliangruber: +1 to avoiding surprises, and functions having side effect is a common surprise. The cached wrapper sounds best to me

backend/lib/pix-service/service.js
```diff
@@ -0,0 +1,44 @@
...
+import { PIECE_INDEXER_URL } from '../config.js'
+import { getMinerPeerId } from '../rpc-service/service.js'
+import { PixResponse } from './data-types.js'
+import { cacheOptions } from './utils.js'
+import pRetry from 'p-retry'
+import { LRUCache } from 'lru-cache'
+
+// A cache for miner peer IDs
+// It has a maximum of 10000 items and deletes the least recently used items
+const minerPeerIdsCache = new LRUCache(cacheOptions)
```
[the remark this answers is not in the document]
NikolasHaimerl: Thanks for opening the issue.

db/migrations/004.do.unique-constraint.sql
```diff
@@ -0,0 +1,13 @@
...
+  activated_at_epoch,
+  miner_id,
+  client_id,
+  piece_cid,
+  piece_size,
+  term_start_epoch,
+  term_min,
+  term_max,
+  sector_id
+);
```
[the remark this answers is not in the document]
NikolasHaimerl: We need the unique constraint to be able to call the `ON CONFLICT` syntax of SQL

backend/test/deal-observer.test.js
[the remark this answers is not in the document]
NikolasHaimerl: You mean whether there is a config option for the pg.Client that automatically converts null to undefined?

backend/lib/config.js
NikolasHaimerl: I created a ticket for this.

backend/bin/deal-observer-backend.js
NikolasHaimerl: I changed the way Sentry is imported

backend/lib/rpc-service/service.js
```diff
@@ -93,3 +108,13 @@ export function getActorEventsFilter (blockHeight, eventTypeString) {
...
   }
 }
+
+/**
+ * @param {number} minerId
+ */
+export function getMinderInfoParameters (minerId) {
+  return [
+    'f0' + minerId.toString(),
+    null
```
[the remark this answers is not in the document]
NikolasHaimerl: I created a ticket for this.

backend/lib/deal-observer.js
```diff
@@ -101,12 +98,18 @@ export async function storeActiveDeals (activeDeals, pgPool) {
...
- * @param {Queryable} pgPool
- * @param {string} query
- * @returns {Promise<Array<Static<typeof ActiveDealDbEntry>>>}
- */
-async function loadDeals (pgPool, query) {
+   * @param {Queryable} pgPool
+   * @param {string} query
+   * @returns {Promise<Array<Static <typeof ActiveDealDbEntry>>>}
+   */
+export async function loadDeals (pgPool, query) {
```
NikolasHaimerl, line 105: We need to export it as we have not finished the PR for extracting the DB functions. https://github.com/CheckerNetwork/deal-observer/pull/39

backend/bin/loops.js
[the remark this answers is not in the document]
NikolasHaimerl: I think a more robust approach would be to simply fetch all payloads that are missing and flag those payloads which re unretrievable as such. Either introduce a new column or set paylod_cid to something like `NON_RETRIEVABLE`. 
This will be part of the second iteration of the piece indexer where missing payloads will be adressed.

db/lib/database-access.js
NikolasHaimerl: The function `fetchDealsWithNoPayloadCid` does specify
`   * @param {number} limit
` wouldn't this deter the kind of SQL injection that you specified?

backend/bin/loops.js
NikolasHaimerl: Now the PR is ready for a full review. I extracted all the unrelated topics into separate PRs.

db/lib/database-access.js
bajtos: Unfortunately not.

`@param {number} limit` is checked at compile time only.

TypeScript has `any` type, which easily sneaks into the codebase unless you enforce strict type checking, which we don't enforce, because it can be tedious in JS/JSDoc codebases.

I believe this would pass the typescript checker:

```ts
/** @type {any} */
const limit = '; DROP TABLE active_deals; --'

await loadDeals(pgPool, 1, limit)
```

Of course, you wouldn't have `@type {any}` in real code, the `any`-typed value would sneak in from a place where we didn't specify the type.

Uncomment the following line in `tsconfig.json` and run `tsc -p .` to see for yourself how many places uses `any`:

https://github.com/CheckerNetwork/deal-observer/blob/92a00b83ee2d52e00a9392e71df154d8bb779de6/tsconfig.json#L16

bajtos: I created a follow-up issue to enable strict type checks: https://github.com/CheckerNetwork/deal-observer/issues/51

backend/bin/deal-observer-backend.js
```diff
@@ -7,6 +7,7 @@ import { createInflux } from '../lib/telemetry.js'
 import { getChainHead, rpcRequest } from '../lib/rpc-service/service.js'
 import { fetchDealWithHighestActivatedEpoch, observeBuiltinActorEvents } from '../lib/deal-observer.js'
 import assert from 'node:assert'
+import { pieceIndexerLoopFunction } from '../lib/piece-indexer.js'
```
juliangruber, line 10:
```suggestion
import { pieceIndexerLoopFunction } from '../lib/piece-indexer.js'
```

Again, `Function` or `Fn` shouldn't be part of a function name, in my opinion. Can we start with a verb instead? Something like `indexPieces()`?

backend/lib/pix-service/service.js
```diff
@@ -0,0 +1,41 @@
...
+      method: 'GET',
+      headers: { 'content-type': 'application/json' }
+    }), { retries: 5 })
+    const json = await response.json()
+    try {
+      const parsedPixResponse = Value.Parse(PixResponse, json)
+      return parsedPixResponse.samples[0]
+    } catch (e) {
+      e.message = `Failed to parse response from piece indexer: ${e.message}, response: ${JSON.stringify(json)}`
+      throw e
```
juliangruber, line 25:
```suggestion
      throw new Error(`Failed to parse response from piece indexer, response: ${JSON.stringify(json)}`, { cause: err })
```

JS has a native construct for error wrapping now :) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause

backend/lib/pix-service/service.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+  } catch (e) {
+    console.error(`Failed to make RPC request: ${e.message}`)
```
juliangruber, line 28:
```suggestion
    console.error('Failed to make RPC request', e)
```

If we only log `e.message`, we lose the stack trace. `Console.error` knows how to treat errors and log them accordingly:

```js
> console.log('hi', new Error('oh no'))
hi Error: oh no
    at REPL1:1:19
    at ContextifyScript.runInThisContext (node:vm:136:12)
    at REPLServer.defaultEval (node:repl:598:22)
    at bound (node:domain:432:15)
    at REPLServer.runBound [as eval] (node:domain:443:12)
    at REPLServer.onLine (node:repl:927:10)
    at REPLServer.emit (node:events:531:35)
    at REPLServer.emit (node:domain:488:12)
    at [_onLine] [as _onLine] (node:internal/readline/interface:417:12)
    at [_line] [as _line] (node:internal/readline/interface:888:18)
```

juliangruber (changes requested):
mostly looking great to me now 👏

db/lib/database-access.js
juliangruber: Great catch Miro!

backend/lib/piece-indexer.js
```diff
@@ -0,0 +1,49 @@
...
+/** @import { Static } from '@sinclair/typebox' */
+/** @import { ActiveDealDbEntry } from '@filecoin-station/deal-observer-db/lib/types.js' */
+
+/**
+ * @param {import("@filecoin-station/deal-observer-db").Queryable} pgPool
+ * @param {function} makeRpcRequest
+ * @param {function} pixRequest
+ * @returns {Promise<void>}
+ */
+export async function updatePayloadCids(pgPool, makeRpcRequest, activeDeals, pixRequest) {
```
pyropy, line 14:
```suggestion
export async function updatePayloadCids(pgPool, makeRpcRequest, activeDeals, makePixRequest) {
```

Name`pixRequest` does not make it obvious that it's a function that makes request to `pix` service.

backend/lib/piece-indexer.js
```diff
@@ -0,0 +1,49 @@
...
+    const updatedDeals = []
+    for (const deal of activeDeals) {
+        const payloadCid = await fetchPayloadCid(deal.miner_id, deal.piece_cid, makeRpcRequest, pixRequest)
```
pyropy, line 17:
```suggestion
        const payloadCid = await fetchPayloadCid(deal.miner_id, deal.piece_cid, makeRpcRequest, makePixRequest)
```

backend/lib/piece-indexer.js
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ *
+ * @param {function} rpcRequest
+ * @param {function} pixRequest
+ * @param {Queryable} pgPool
+ * @param {number} queryLimit
+ * @returns {Promise<void>}
+ */
+export const pieceIndexerLoopFunction = async (rpcRequest, pixRequest, pgPool, queryLimit) => {
```
pyropy, line 32:
```suggestion
export const pieceIndexerLoopFunction = async (makeRpcRequest, makePixRequest, pgPool, queryLimit) => {
```

Renaming `rpcRequest` to `makeRpcRequest` would make things a bit more consistent.

backend/lib/piece-indexer.js
```diff
@@ -0,0 +1,49 @@
...
+    // TODO: handle payloads which cannot be retrieved from the piece CID indexer
+    const dealsWithMissingPayloadCid = await fetchDealsWithNoPayloadCid(pgPool, queryLimit)
+    if (dealsWithMissingPayloadCid !== null && dealsWithMissingPayloadCid) {
+        await updatePayloadCids(pgPool, rpcRequest, dealsWithMissingPayloadCid, pixRequest)
```
pyropy, line 36:
```suggestion
        await updatePayloadCids(pgPool, makeRpcRequest, dealsWithMissingPayloadCid, makePRequest)
```

backend/lib/piece-indexer.js
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/**
+   * @param {Queryable} pgPool
+   * @param {number} limit
+   * @returns {Promise<Array<Static< typeof ActiveDealDbEntry>>>}
+   */
+export async function fetchDealsWithNoPayloadCid(pgPool, limit) {
+    const query = `SELECT * FROM active_deals WHERE payload_cid IS NULL ORDER BY activated_at_epoch ASC LIMIT ${limit}`
+    const result = await parseDeals(pgPool, query)
```
pyropy, line 47:
```suggestion
    return await parseDeals(pgPool, query)
```

backend/lib/pix-service/service.js
```diff
@@ -0,0 +1,41 @@
...
+import { getMinerPeerId } from '../rpc-service/service.js'
+import { PixResponse } from './data-types.js'
+import pRetry from 'p-retry'
+
+/**
+ * @param {string} providerId
+ * @param {string} pieceCid
+ * @returns {Promise<string>}
+ */
+export const pixRequest = async (providerId, pieceCid) => {
```
pyropy, line 12: Maybe naming this function something like `getDealPayloadCid` would make more sense as `pixRequest` is too generic, what do you think?

pyropy (commented):
Great work, I think we're almost there.

backend/bin/deal-observer-backend.js
NikolasHaimerl: I renamed the function.

backend/lib/pix-service/service.js
NikolasHaimerl: Thanks you!

db/migrations/004.do.unique-constraint.sql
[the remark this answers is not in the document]
bajtos: > We need the unique constraint to be able to call the `ON CONFLICT` syntax of SQL

I see!

I am happy to introduce a UNIQUE constraint, but I'd like to ensure we have enough understanding to pick the right combination of columns.

For example, in https://github.com/CheckerNetwork/deal-observer/pull/33, @pyropy uses the following tuple: `(miner_id, client_id, piece_cid, piece_size)`, see https://github.com/CheckerNetwork/deal-observer/pull/33#discussion_r1935879448

I don't know enough myself to provide an authoritative answer. I feel that `sector_id` should be most likely included. I think `term_*` are tied to a deal and are not part of the UNIQUE tuple. I am not sure about `activated_at_epoch`.

I am not saying we must use the same index and/or UNIQUE constraint for your work and #33; I just want us to find a solution that considers both angles.

**On the second thought, I am okay to use the UNIQUE index you proposed here. Let's ship what you have here, keep learning more about the space after this PR is landed, possibly by inspecting the data we have in our production DB.**

backend/lib/piece-indexer.js
```diff
@@ -0,0 +1,48 @@
...
+ * @returns {Promise<void>}
+ */
+export async function updatePayloadCids (pgPool, makeRpcRequest, activeDeals, pixRequest) {
+  const updatedDeals = []
+  for (const deal of activeDeals) {
+    const payloadCid = await fetchPayloadCid(deal.miner_id, deal.piece_cid, makeRpcRequest, pixRequest)
+    deal.payload_cid = payloadCid
+    updatedDeals.push(deal)
+  }
+  await storeActiveDeals(updatedDeals, pgPool)
```
bajtos, line 22: While re-inserting deals to update one field may work, it's not the usual way how to implement updates in SQL. It will likely conflict with other updates, e.g. the one @pyropy is implementing in https://github.com/CheckerNetwork/deal-observer/pull/33

Let's rework the code in this pull request to use `UPDATE` statement to update the payload CIDs.

Since we are mapping `(miner_id, piece_cid)` to `payload_cid`, the query could look like this:

```js
await pgPool.query(`
  UPDATE active_deals
  SET payload_cid = $1
  WHERE miner_id = $2 AND piece_cid = $3
`, [
  deal.payload_cid,
  deal.miner_id,
  deal.piece_cid
])
```

You can run this inside the for loop, right after you fetched the payload CID.

As a side effect, if `fetchPayloadCid` throws an error, the deals we resolved before the error are still updated.

A more efficient solution is to do a bulk update, but I don't know the SQL syntax from the top of my head. @pyropy will need it in #33; you can reuse his solution.

For now, I think it's ok to do one SQL query per deal, because we are processing deals sequentially and waiting for an HTTP request to complete for each deal, so the number of SQL queries is not a bottleneck.

backend/lib/piece-indexer.js
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * @param {import("@filecoin-station/deal-observer-db").Queryable} pgPool
+ * @param {function} makeRpcRequest
+ * @param {function} pixRequest
+ * @returns {Promise<void>}
+ */
+export async function updatePayloadCids (pgPool, makeRpcRequest, activeDeals, pixRequest) {
+  const updatedDeals = []
+  for (const deal of activeDeals) {
+    const payloadCid = await fetchPayloadCid(deal.miner_id, deal.piece_cid, makeRpcRequest, pixRequest)
```
bajtos, line 18: There may be multiple deals with the same combination of `(miner_id, piece_cid)` values, that's how the currently proposed UNIQUE constraint works.

From what I've seen in fil-deal-ingester, it's very very likely there will be many such duplicates.

I created a follow-up issue to improve a more efficient handling of such duplicates, see https://github.com/CheckerNetwork/deal-observer/issues/59

bajtos (changes requested):

backend/lib/piece-indexer.js
NikolasHaimerl: I see, is there a certain reason why we are doing it in [spark-evalute](https://github.com/CheckerNetwork/spark-evaluate/blob/29527a07c30fee203b68e2852f0365ac79b46b39/lib/public-stats.js#L50) then?

NikolasHaimerl: I created an issue to speed up the SQL query.

bajtos: In spark evaluate, we update an aggregate number by adding a new value to the existing value. We use INSERT to handle the initial case when there is no existing value.

There are also around 1000 updates at once in spark-evaluate, it would take too long to run one query for each update. Our situation here in deal-observer is different because we make an HTTP request before updating each deal.

backend/bin/loops.js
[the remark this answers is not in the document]
bajtos: I believe https://github.com/CheckerNetwork/spark-deal-observer/issues/30 is tracking the follow-up work on this, right? If yes, then I think we can resolve this thread. WDYT?

backend/bin/deal-observer-backend.js
```diff
@@ -63,6 +63,32 @@ const dealObserverLoop = async (makeRpcRequest, pgPool) => {
   }
 }
 
+export const pieceIndexerLoop = async (makeRpcRequest, makePixRequest, pgPool) => {
```
juliangruber, line 66: I believe this function is never called

backend/lib/deal-observer.js
```diff
@@ -89,34 +74,37 @@ export async function storeActiveDeals (activeDeals, pgPool) {
...
- * @param {Queryable} pgPool
- * @param {string} query
- * @returns {Promise<Array<Static<typeof ActiveDealDbEntry>>>}
- */
-async function loadDeals (pgPool, query) {
+   * @param {Queryable} pgPool
+   * @param {string} query
+   * @returns {Promise<Array<Static <typeof ActiveDealDbEntry>>>}
+   */
+export async function parseDeals (pgPool, query) {
```
pyropy, line 100: I have mixed feelings about this functions, it's called `parseDeals` but it queries and parses deals. Maybe it should just parse already queried deals?

backend/bin/deal-observer-backend.js
```diff
@@ -23,8 +24,7 @@ if (!INFLUXDB_TOKEN) {
 assert(SPARK_API_BASE_URL, 'SPARK_API_BASE_URL required')
 assert(SPARK_API_TOKEN, 'SPARK_API_TOKEN required')
 
-const OBSERVE_ACTOR_EVENTS_LOOP_INTERVAL = 10 * 1000
-const SPARK_API_SUBMIT_DEALS_LOOP_INTERVAL = 10 * 1000
+const LOOP_INTERVAL = 10 * 1000
```
bajtos, line 27: I think we should keep two constants, because we are likely to use different loop delay value for each loop. We already discussed changing `OBSERVE_ACTOR_EVENTS_LOOP_INTERVAL` from 10s to something like 30s or 100s, see https://github.com/CheckerNetwork/spark-deal-observer/pull/29#discussion_r1931754959

Having one constant will make such changes slightly more difficult to implement in the future.

This comment is not blocking this PR from being landed, though.

backend/bin/deal-observer-backend.js
bajtos: IIRC, @NikolasHaimerl mentioned in an older comment that he does not want to run it yet, but I don't remember details.

I see you already added a call of this function at the bottom of this file, so I think this comment is resolved now?

```js
await Promise.all([
  // TODO: Define `pixRequest`
  pieceIndexerLoop(rpcRequest, () => {}, pgPool),
  observeActorEventsLoop(rpcRequest, pgPool),
  sparkApiSubmitDealsLoop(pgPool, {
    sparkApiBaseUrl: SPARK_API_BASE_URL,
    // ...
  })
])
```

backend/lib/deal-observer.js
```diff
@@ -15,10 +14,8 @@ import { Value } from '@sinclair/typebox/value'
  */
 export async function observeBuiltinActorEvents (blockHeight, pgPool, makeRpcRequest) {
   const eventType = 'claim'
-  const activeDeals = await getActorEvents(getActorEventsFilter(blockHeight, eventType), makeRpcRequest)
-  assert(activeDeals !== undefined, `No ${eventType} events found in block ${blockHeight}`)
-  console.log(`Observed ${activeDeals.length} ${eventType} events in block ${blockHeight}`)
```
bajtos, line 20: Can we please preserve this line? I found it useful to see in the logs how many deals we are adding per block and which loop iterations don't find any new blocks.

Again, this is not a blocker.

backend/lib/deal-observer.js
bajtos: I believe we agreed to call it `loadDeals`, I am not sure why it's being renamed back to `parseDeals`.

backend/lib/piece-indexer.js
```diff
@@ -0,0 +1,82 @@
...
+ * @returns {Promise<void>}
+ */
+export async function updatePayloadCids (pgPool, makeRpcRequest, activeDeals, makePixRequest) {
+  const updatedDeals = []
+  for (const deal of activeDeals) {
+    const payloadCid = await fetchPayloadCid(deal.miner_id, deal.piece_cid, makeRpcRequest, makePixRequest)
+    deal.payload_cid = payloadCid
+    updatedDeals.push(deal)
+  }
+  await updatePayloadInActiveDeal(pgPool, updatedDeals)
```
bajtos, line 23: Since this is updating multiple deals at once, the name should be plural.

```suggestion
  await updatePayloadInActiveDeals(pgPool, updatedDeals)
```

bajtos, line 23: This is not great:

- We loop over all deals to update the in-memory deal objects with PayloadCID.
- Then inside `updatePayloadInActiveDeals`, we loop through those deals again to run one SQL query per deal.

Wouldn't it be better to have only one loop?

```js
  const updatedDeals = []
  for (const deal of activeDeals) {
    const payloadCid = await fetchPayloadCid(deal.miner_id, deal.piece_cid, makeRpcRequest, makePixRequest)
    deal.payload_cid = payloadCid
    await updatePayloadInActiveDeal(pgPool, updatedDeals)
  }
```

This is especially important in case `fetchPayloadCid` throws an error after we already resolved some payload CIDs. The current implementation will discard (not persist) the already resolved payload CIDs.

backend/lib/pix-service/service.js
```diff
@@ -0,0 +1,43 @@
...
+    const response = await pRetry(async () => await fetch(url, {
+      method: 'GET',
+      headers: { 'content-type': 'application/json' }
+    }), { retries: 5 })
+    const json = await response.json()
+    try {
+      const parsedPixResponse = Value.Parse(PixResponse, json)
+      if (parsedPixResponse.samples.length === 0) {
+        throw new Error('No samples found in the response.')
+      }
```
bajtos, line 24: This is expected, it will happen often (in ~60% of calls).

I propose that `getDealPayloadCid()` should return `undefined` when this happens.

That would also help with addressing the comment above:

```
export const indexPieces = async (makeRpcRequest, makePixRequest, pgPool, queryLimit) => {
  // TODO: handle payloads which cannot be retrieved from the piece CID indexer
```

bajtos (commented):
I reviewed the implementation, not looking much at the test files. The changes looks reasonable. I left some comments to consider above.

The only thing I consider as required to fix is how we handle the case where the PieceIndexer returns an empty list of PayloadCIDs.

backend/bin/deal-observer-backend.js
juliangruber: just need to fix `pixRequest`, but yes 👍

backend/bin/deal-observer-backend.js
```diff
@@ -107,17 +108,45 @@ const sparkApiSubmitDealsLoop = async (pgPool, { sparkApiBaseUrl, sparkApiToken,
...
+    }
+    if (dt < LOOP_INTERVAL) {
+      await timers.setTimeout(LOOP_INTERVAL - dt)
     }
   }
 }
 
 await Promise.all([
+  // TODO: Define `pixRequest`
+  pieceIndexerLoop(rpcRequest, () => {}, pgPool),
```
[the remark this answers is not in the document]
juliangruber: The implementation of `pixRequest` so far is missing from this PR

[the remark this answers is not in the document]
bajtos: I think you `getDealPayloadCid` is the implementation of `pixRequest`.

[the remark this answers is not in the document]
bajtos: Here is the code invoking `pixRequest`:

```js
export async function fetchPayloadCid (providerId, pieceCid, makeRpcRequest, makePixRequest) {
  const minerPeerId = await getMinerPeerId(providerId, makeRpcRequest)
  const payloadCid = await makePixRequest(minerPeerId, pieceCid)
  return payloadCid
}
```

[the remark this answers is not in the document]
bajtos: And here is how is `getDealPayloadCid` defined:

```js
export const getDealPayloadCid = async (providerId, pieceCid) => {
```

It looks like a match to me.

backend/lib/pix-service/service.js
```diff
@@ -0,0 +1,43 @@
...
+import { getMinerPeerId } from '../rpc-service/service.js'
+import { PixResponse } from './data-types.js'
+import pRetry from 'p-retry'
+
+/**
+ * @param {string} providerId
+ * @param {string} pieceCid
+ * @returns {Promise<string>}
+ */
+export const getDealPayloadCid = async (providerId, pieceCid) => {
```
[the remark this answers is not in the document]
bajtos: See https://github.com/CheckerNetwork/spark-deal-observer/pull/31#discussion_r1937622057

backend/test/deal-observer.test.js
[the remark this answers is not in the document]
juliangruber: I think @bajtos meant to check if the Postgres lib returns `undefined` or `null`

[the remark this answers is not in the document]
juliangruber: Since if it returned `null`, we would have a problem with TypeBox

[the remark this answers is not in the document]
juliangruber: I'm assuming for now that tests will cover this, otherwise will follow up with a fix

backend/bin/deal-observer-backend.js
```diff
@@ -108,20 +109,48 @@ const sparkApiSubmitDealsLoop = async (pgPool, { sparkApiBaseUrl, sparkApiToken,
...
+      await timers.setTimeout(LOOP_INTERVAL - dt)
+    }
+  }
+}
+
+export const pieceIndexerLoop = async (makeRpcRequest, makePixRequest, pgPool) => {
+  const LOOP_NAME = 'Piece Indexer'
+  while (true) {
+    const start = Date.now()
+    const queryLimit = 1000
```
juliangruber, line 129: How did we come up with this magic number? It defines how many active deals we look at at a time. It looks like by tuning it we control the trade off between performance and memory usage (but not concurrency).

bajtos: I don't remember the details; it's best to ask @NikolasHaimerl. Your description looks correct to me.

This will be easy to tweak later, I think we should pick some reasonable number for now and move on. WDYT?

bajtos (approved):
I quickly skimmed throught the latest version and it looks very reasonable to me. I don't see any major issues/blockers.

backend/lib/piece-indexer.js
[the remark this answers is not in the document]
juliangruber: Good call, I found that confusing as well. Will fix

PR #33, opened 2025-01-28
Deal submissions

### LINKED ISSUE #26: Periodically add new deals to spark-api's database of eligible deals
Implement a new loop to periodically check the `active_deals` table for new deals that have `payload_cid` set and submit these new deals to spark-api using the recently added endpoint `POST /eligible-deals-batch` (see https://github.com/filecoin-station/spark-api/pull/502).

This work requires two tasks to be finished first:
- https://github.com/filecoin-station/deal-observer/pull/12 - the loop populating `active_deals` from built-in actor events
- a loop to query PieceIndexer to fill `payload_cid`

Notes:

(1)
f05 deals ingested via [fil-deal-ingester](https://github.com/filecoin-station/fil-deal-ingester) have priority over deals discovered from built-in actor events + PieceIndexer. This is important to preserve backwards compatibility. The Payload CID returned by PieceIndexer can differ from the Payload CID found in f05 `DealProposal.Label` field.

deal-observer is designed in such a way that it does not distinguish f05 and DDO deals. That's intentional, as it's the design we need for Spark v2.

We run fil-deal-ingester only once a day, and the process takes up to 6-8 hours to complete. On the other hand, deal-observer ingests new (f05) deals with a delay of 940 epochs (~8 hours). If we submit new deals to spark-api as soon as we discover them, then we will front-run fil-deal-ingester and create entries in the database of eligible deals with a different Payload CID.

To avoid that, we need to delay submitting new deals to spark-api long enough for the fil-deal-ingester service to ingest f05 deals first.

Let's create a healthy safety margin and set the delay to 48 hours.

(2)
Handling the expiration of deals using the built-in actors' events is tricky. To keep things simple and ship the first iteration soon, let's use the following formula to calculate the `expires_at` field required by spark-api:

(Quoting from https://github.com/space-meridian/roadmap/issues/172#issuecomment-2573450933):
> Calculate the deal expiration time (`expires_at`) as `term-start` (epoch) + `term-min` (duration in epochs). This will often be inaccurate, and we will stop testing deals that are still active, but it makes it easy for us to start somewhere.

(3)
deal-observer should submit multiple deals in one request with a reasonable batch size. spark-api limits the request body size to 100MB.

(4)
deal-observer will need to keep track of deals that were successfully submitted to spark-api. I propose adding a new column to `active_deals` table. This can be a boolean flag, but maybe a timestamp would be more useful if we need to troubleshoot.

TODO

- [ ] Address review comments in https://github.com/CheckerNetwork/deal-observer/pull/33
- [ ] Deploy to production (first week of Feb)

pyropy:
Introduces new loop for deal submission. Deal submission loop periodically queries database for unsubmitted deals, formats them and submits to spark api in batches. Deals are considered for submission if they were created more than two days ago, were not submitted previously (`submitted_at` timestamp field is null), have payload CID defined and are not yet expired.

Closes #26


backend/bin/deal-observer-backend.js
```diff
@@ -40,21 +52,45 @@ const dealObserverLoop = async (makeRpcRequest, pgPool) => {
...
+        point.intField('duration_ms', dt)
+      })
+    }
+    if (dt < DEAL_OBSERVER_LOOP_INTERVAL) {
+      await timers.setTimeout(DEAL_OBSERVER_LOOP_INTERVAL - dt)
+    }
+  }
+}
+
+const dealSubmitterLoop = async (pgPool, sparkApiBaseURL, dealIngestionAccessToken) => {
```
pyropy, line 69: I am thinking about moving this to separate file but this will make deployment harder.

db/migrations/004.do.active-deals-timestamps.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE active_deals 
+    ADD COLUMN IF NOT EXISTS submitted_at TIMESTAMP,
+    ADD COLUMN IF NOT EXISTS created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP;
```
pyropy, line 3: If would consider replacing `created_at` with `term_start_epoch` If that's possible? How often will`term_start_epoch` be set to some date in the past if we're listening for real-time events?

backend/lib/deal-submitter.js
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Submits eligible deals to a spark api.
+ *
+ * @param {string} sparkApiBaseURL
+ * @param {string} dealIngestionAccessToken
+ * @param {Array} eligibleDeals
+ * @returns {Promise<void>}
+ */
+export const submitEligibleDeals = async (sparkApiBaseURL, dealIngestionAccessToken, eligibleDeals) => {
```
pyropy, line 102: I was considering currying this function, so signature would be something like:

```
const submitEligibleDeals = (sparkApiBaseURL, dealIngestionAccessToken) => async (eligibleDeals) => Promise<void>
```

In the case we curry this function then we would not have to pass `sparkApiBaseURL` and  `dealIngestionAccessToken` to `findAndSubmitEligibleDeals` as arguments

backend/bin/deal-observer-backend.js
bajtos: @pyropy please take a look at the discussion in https://github.com/filecoin-station/deal-observer/pull/42 where we are discussing how to refactor the code to make it easier to add more loops (IPNI lookup, your submission to spark-api, etc.)

db/migrations/004.do.active-deals-timestamps.sql
bajtos: I'd like the database to store the raw data we are getting from the Filecoin on-chain state, and calculate the timestamps from that. The algorithm for calculating the timestamps can change in the future as we gain more understanding of the deal lifecycle and actor events. By keeping the raw data in our table, it's easy to change the way how we interpret it.

To find the time when a deal was created, you should look at `activated_at_epoch` and convert it to a timestamp.

db/migrations/005.do.epoch-to-timestamp.sql
```diff
@@ -0,0 +1,9 @@
+CREATE OR REPLACE FUNCTION epoch_to_timestamp(epoch INTEGER)
+RETURNS TIMESTAMP AS $$
+DECLARE
+    FILECOIN_GENESIS_UNIX_EPOCH CONSTANT INTEGER := 1598306400;
+BEGIN
+    -- Calculate and return the UNIX timestamp
+    RETURN TO_TIMESTAMP(epoch * 30 + FILECOIN_GENESIS_UNIX_EPOCH);
+END;
+$$ LANGUAGE plpgsql;
```
bajtos, line 9: Nice trick to define a stored procedure for converting epochs to timestamps.

Most likely, we will need to add indices (indexes) to speed up the query looking up deals to submit. Do you know if it's possible to index over values calculated using a stored procedure?

Something like this:

```sql
CREATE INDEX active_deals_created_at ON active_deals (epoch_to_timestamp(activated_at_epoch))
```

Maybe should convert timestamps to epochs in the query `WHERE` component, WDYT?

```sql
-- migration script
CREATE INDEX active_deals_activated_at_epoch ON active_deals (activated_at_epoch);
-- query
SELECT 
 -- ...
FROM active_deals
WHERE
  activated_at_epoch < timestamp_to_epoch(NOW() - INTERVAL '2 days')
```

backend/lib/deal-submitter.js
```diff
@@ -0,0 +1,112 @@
...
+ * Mark eligible deals as submitted.
+ *
+ * @param {Queryable} pgPool
+ * @param {Array} eligibleDeals
+ */
+const markEligibleDealsSubmitted = async (pgPool, eligibleDeals) => {
+  await pgPool.query(`
+    UPDATE active_deals
+    SET submitted_at = NOW()
+    WHERE miner_id = ANY($1) AND client_id = ANY($2) AND piece_cid = ANY($3) AND payload_cid = ANY($4)
```
bajtos, line 86: It's important to include `piece_size` in this key. The Filecoin core devs keep insisting that PieceCID is not enough to uniquely identify a piece, you need both PieceCID + PieceSize.

I think it's safe to drop `payload_cid`, because that's calculated from `miner_id` and `piece_cid`.

bajtos:
@pyropy Great start!

Please sync with @NikolasHaimerl on how to run three loops inside the backend process - we have the first loop in place, you are adding a new loop and Nikolas is adding a new loop too. (See also https://github.com/filecoin-station/deal-observer/pull/42.)

How is your implementation going to handle errors? In particular, if the HTTP request to spark-api fails.

The spark-api implementation makes it safe to re-submit the same deals again, but I'd rather avoid resubmitting thousands of deals every time an HTTP request to spark-api fail.

db/migrations/004.do.active-deals-timestamps.sql
pyropy: Thank you for this info! I've dropped `created_at` column 🙏🏻

db/migrations/005.do.epoch-to-timestamp.sql
pyropy: Sadly I don't think it's possible to create index over a calculated value like that.

I do like the second option of introducing `timestamp_to_epoch` procedure and adding index to `activated_at_epoch` field.

backend/bin/deal-observer-backend.js
```diff
@@ -6,21 +7,33 @@ import '../lib/instrument.js'
...
-const { INFLUXDB_TOKEN } = process.env
 if (!INFLUXDB_TOKEN) {
   console.error('INFLUXDB_TOKEN not provided. Telemetry will not be recorded.')
 }
-const LOOP_INTERVAL = 10 * 1000
+assert(SPARK_API_BASE_URL, 'SPARK_API_BASE_URL required')
+assert(DEAL_INGESTER_TOKEN, 'DEAL_INGESTER_TOKEN required')
+
+const DEAL_OBSERVER_LOOP_INTERVAL = 10 * 1000
+const DEAL_SUBMITTER_LOOP_INTERVAL = 10 * 1000
```
juliangruber, line 26: It took me a bit to understand the meaning of "deal observer" and "deal submitter". At first I thought, are we the ones submitting the deals, that the observer then observes? After looking at the comments in deal-submitter.js I'm realizing that we're submitting deals to spark-api.

Can we find a more self explanatory pair of names? Deal observer is relatively clear - it observes deals. Deal submitter isn't optimal.

Here are some options I'm thinking of:

- deal syncer
- spark api deal syncer

Maybe we could also add "eligible" (although I personally am at war with this word):

- eligible deal syncer

For my understanding: Why is it that observing and submitting/sharing/syncing deals are two steps? Could it also make sense to submit/share/sync deals as we observe them, directly?

backend/lib/deal-submitter.js
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Find deals that are eligible to be submitted to a spark api and submits them.
+ *
+ * @param {PgPool} pgPool
+ * @param {string} sparkApiBaseURL
+ * @param {string} dealIngestionAccessToken
+ * @param {number} batchSize
+ * @param {(sparkApiBaseURL: string, dealIngestionAccessToken: string, eligibleDeals: Array) => Promise<void>} submitEligibleDealsFn
+ */
+export const findAndSubmitEligibleDeals = async (pgPool, sparkApiBaseURL, dealIngestionAccessToken, batchSize, submitEligibleDealsFn) => {
```
juliangruber, line 13: This argument list is quite long for positional arguments.

@bajtos what do you think about also adding this to the linter? Something like positional arguments must be 3 or less?

backend/lib/deal-submitter.js
```diff
@@ -0,0 +1,115 @@
...
+  client.release()
+}
+
+/**
+ * Format eligible deals to format expected by spark api.
+ *
+ * @param {Array} deals
+ * @returns {Array}
+*/
+const formatEligibleDeals = (deals) => {
```
juliangruber, line 68: `format` is a bit vague here, what about `formatEligibleDealsForSparkApi`?

backend/lib/deal-submitter.js
```diff
@@ -0,0 +1,115 @@
...
+ * @param {PgPool} pgPool
+ * @param {string} sparkApiBaseURL
+ * @param {string} dealIngestionAccessToken
+ * @param {number} batchSize
+ * @param {(sparkApiBaseURL: string, dealIngestionAccessToken: string, eligibleDeals: Array) => Promise<void>} submitEligibleDealsFn
+ */
+export const findAndSubmitEligibleDeals = async (pgPool, sparkApiBaseURL, dealIngestionAccessToken, batchSize, submitEligibleDealsFn) => {
+  for await (const eligibleDeals of findEligibleDeals(pgPool, batchSize)) {
+    const formattedEligibleDeals = formatEligibleDeals(eligibleDeals)
+    await submitEligibleDealsFn(sparkApiBaseURL, dealIngestionAccessToken, formattedEligibleDeals)
```
juliangruber, line 16: The `Fn` suffix is redundant imo, since the variable name already starts with a verb. What do you think about just `submitEligibleDeals`?

backend/lib/deal-submitter.js
```diff
@@ -0,0 +1,115 @@
...
+    await markEligibleDealsSubmitted(pgPool, eligibleDeals)
```
juliangruber, line 17: What about `markEligibleDealsAsSubmitted`? At first I thought this means to mark the submitted eligible deals.

backend/lib/deal-submitter.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+}
+
+/**
+ * Find deals that are eligible to be submitted to a spark api.
```
juliangruber, line 22:
```suggestion
 * Find deals that are eligible to be submitted to spark api.
```

There is only one spark api

backend/lib/deal-submitter.js
```diff
@@ -0,0 +1,115 @@
...
+ * Eligible deals are those that have not been submitted yet,
+ * were created more than 2 days ago, have payload cid, and have not yet expired.
```
juliangruber, line 24:
```suggestion
 * Eligible deals are those that
 * - have not been submitted yet
 * - were created more than 2 days ago
 * - have payload cid
 * - have not yet expired
```

juliangruber (changes requested):

backend/lib/epoch.js
```diff
@@ -0,0 +1,21 @@
+// Copied from: https://observablehq.com/@protocol/filecoin-epoch-calculator
```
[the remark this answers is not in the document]
pyropy: This notebook is not licensed by its author. Here's a quote from [ObservableHQ docs](https://observablehq.com/documentation/collaboration/licenses) about using ObservableHQ notebooks:

> Observable’s [terms of service](https://observablehq.com/terms-of-service#5-license-grant-to-other-users) allow authors to fork and edit published notebooks here on Observable, but beyond that it’s up to the notebook authors to decide what kind of use they want to allow.

I've dug into the Terms of Service and under section 5. License Grant to Other Users and they say the following: 

> Any User-Generated Content you or your Team Users post publicly may be viewed by others. By publishing your notebooks, you agree to allow others to view, “import” and “fork” your notebooks. This means that others may load your published Content into notebooks they control, or make their own copies of your published Content in notebooks they control.
>
> When you publish a notebook, you grant each User of Observable a nonexclusive, worldwide license to use, display, and perform Your Content through the Observable Service and to reproduce Your Content solely on Observable as permitted through Observable’s functionality (for example, through forking). If you are uploading Content you did not create or own, you are responsible for ensuring that the Content you upload is licensed under terms that grant these permissions to other Observable Users.
>
> This license does not grant Other Users the right to sell Your Content or otherwise distribute or use it outside of our provision of the Service.

 If we fork this dashboard and grant license we should be okay? Also this code is not exact copy and API is a bit different. Let me know what you think.

backend/lib/deal-submitter.js
pyropy: I think you're right on this. Maybe we should drop "eligible" altogether and just go with "unsubmitted", "unposted" or "unsynced" so we would end up with functions name like `findUnsubmittedDeals`, `findUnpostedDeals` or `findUnsyncedDeals`.

backend/bin/deal-observer-backend.js
pyropy: I agree, "deal submitter" seems kind of vague, a more descriptive name would be better. IMO just changing the verb from "submit" to "sync" would not add more information about the exact operation, but using something like "spark api deal submitter" or "spark api deal syncer" would add more context of the operation being done. 

I like the idea of "sync" and maybe also we could maybe use other verbs like "post" and "ingest".

> For my understanding: Why is it that observing and submitting/sharing/syncing deals are two steps? Could it also make sense to submit/share/sync deals as we observe them, directly?

@bajtos has described this problem in #26 and the reason we are adding this "delay"
> If we submit new deals to spark-api as soon as we discover them, then we will front-run fil-deal-ingester and create entries in the database of eligible deals with a different Payload CID.

juliangruber: >If we submit new deals to spark-api as soon as we discover them, then we will front-run fil-deal-ingester and create entries in the database of eligible deals with a different Payload CID.


Is there still a chance of this unintended behavior to happen? If fil-deal-ingester should run first, can we ensure that in a more reliable way?

backend/lib/epoch.js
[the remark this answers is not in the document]
juliangruber: Thank you so much for researching this! ⭐ 

> and to reproduce Your Content solely on Observable as permitted through Observable’s functionality (for example, through forking)
> This license does not grant Other Users the right to sell Your Content or otherwise distribute or use it outside of our provision of the Service.

Looks like by default we aren't allowed to use it.

We should be safe, especially since Oli is also from the PL ecosystem. I just wanted to check. I think we're good here

backend/lib/deal-submitter.js
pyropy: I have made this shorter now by currying `submitDealsToSparkApi` function. Now `findAndSubmitDeals` functions accepts three arguments: pgPool, batchSize, and submitDeals.

backend/bin/deal-observer-backend.js
pyropy: I have reached out to @NikolasHaimerl and also left a [comment](https://github.com/CheckerNetwork/deal-observer/pull/42/files#r1933861920). Maye we can discuss it there?

backend/bin/deal-observer-backend.js
pyropy: > Is there still a chance of this unintended behavior to happen? If fil-deal-ingester should run first, can we ensure that in a more reliable ?

I don't think there's an easy way to avoid this architecture until we sunset `fil-deal-ingester`. One way to have both `deal-observer` and`fil-deal-ingester` would be for them to share some lock via external service like postgres. But in return this would make things even more complex and in the end would not remove this additional loop (while `fil-deal-ingester` holds the lock we can't report new deals, hence we have to have some loop that will still submit deals).

backend/bin/deal-observer-backend.js
pyropy: I'm going to keep old way of structuring loops until #42 gets merged. If #42 gets merged before I am going to pull changes and refactor my code accordingly.

db/migrations/004.do.active-deals-submitted-at.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE active_deals ADD COLUMN IF NOT EXISTS submitted_at TIMESTAMP;
+
+CREATE INDEX IF NOT EXISTS active_deals_activated_at_epoch ON active_deals (activated_at_epoch);
```
pyropy, line 3: I am wondering if I should move this to separate migration file?

backend/bin/deal-observer-backend.js
```diff
@@ -40,21 +54,59 @@ const dealObserverLoop = async (makeRpcRequest, pgPool) => {
...
+ * Periodically fetches unsubmitted deals from the database and submits them to the Spark API.
+ *
+ * @param {PgPool} pgPool
+ * @param {object} args
+ * @param {string} args.sparkApiBaseUrl
+ * @param {string} args.dealIngestionAccessToken
+ * @param {number} args.dealSubmitterBatchSize
+ */
+const sparkApiDealSubmitterLoop = async (pgPool, { sparkApiBaseUrl, dealIngestionAccessToken, dealSubmitterBatchSize }) => {
+  const submitDeals = submitDealsToSparkApi(sparkApiBaseUrl, dealIngestionAccessToken)
```
bajtos, line 81: I like the idea of having a `submitDeals` callback that accepts only `unsubmittedDeals` 👍🏻 

However, I expect a function called `submitDealsToSparkApi` to submit the deals, not to create another function.

Can we improve the name to make this clear, please? For example, we can prefix the function name with `create` or add a `Factory` suffix to the name.

For example:

```suggestion
  const submitDeals = createFnSubmitDealsToSparkApi(sparkApiBaseUrl, dealIngestionAccessToken)
```

Few more alternatives:
- createSubmitDealsToSparkApiCallback
- createSubmitDealsToSparkApiCb

If we cannot find a good name, then I think it's a sign we are creating a wrong abstraction.

An alternative solution to consider: curry the function here.

```js
const submitDeals = submitDealsToSparkApi(unsubmittedDeals, sparkApiBaseUrl, dealIngestionAccessToken)
```

Or even this:

```js
await findAndSubmitUnsubmittedDeals(
  pgPool, 
  dealSubmitterBatchSize,
  (unsubmittedDeals) => submitDealsToSparkApi(unsubmittedDeals, sparkApiBaseUrl, dealIngestionAccessToken)
)

File: backend/lib/epoch.js (hunk 13 above)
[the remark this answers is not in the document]
Comment (bajtos) [reply backend/lib/epoch.js:1]: It's great that you remembered to check for the license!

How about reusing this snippet we already have in our codebase?

https://github.com/CheckerNetwork/fil-deal-ingester/blob/a185d7d0038b3f220d4c3c4ff20cbddc7759c376/scripts/parse-retrievable-deals.js#L8-L9

```
// See https://docs.filecoin.io/networks/mainnet#genesis
const GENESIS_TS = new Date('2020-08-24T22:00:00Z').getTime()
```

Used also here:
https://github.com/CheckerNetwork/spark/blob/faf184b267fe29a7e27e500f84a4f5cb717560c0/lib/drand-client.js#L7-L8

File: backend/lib/spark-api-deal-submitter.js [hunk 16]
```diff
@@ -0,0 +1,127 @@
...
+  console.debug(`Finding and submitting deals using batchSize: ${batchSize}`)
+  for await (const unsubmittedDeals of findUnsubmittedDeals(pgPool, batchSize)) {
+    console.debug(`Found ${unsubmittedDeals.length} unsubmitted deals`)
+    try {
+      const formattedDeals = formatUnsubmittedDealsForSparkApi(unsubmittedDeals)
+      await submitDeals(formattedDeals)
+      console.debug(`Successfully submitted ${formattedDeals.length} deals`)
+      await markDealsAsSubmitted(pgPool, unsubmittedDeals)
+    } catch (e) {
+      console.error('Failed to submit deals:', e)
```
bajtos, line 21: Should we report the error to Sentry?

backend/lib/spark-api-deal-submitter.js
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+}
+
+/**
+ * Finds deals that haven't been submitted to spark api.
+ * For deals to be submitted they must
+ * - have not been submitted yet
+ * - were created more than 2 days ago
```
bajtos, line 30: Can you please briefly explain why we submit only days older than 2 days and/or link to the GitHub comment explaining the details.

backend/lib/spark-api-deal-submitter.js
```diff
@@ -0,0 +1,127 @@
...
+ * Mark deals as submitted.
+ *
+ * @param {Queryable} pgPool
+ * @param {Array} eligibleDeals
+ */
+const markDealsAsSubmitted = async (pgPool, eligibleDeals) => {
+  await pgPool.query(`
+    UPDATE active_deals
+    SET submitted_at = NOW()
+    WHERE miner_id = ANY($1) AND client_id = ANY($2) AND piece_cid = ANY($3) AND piece_size = ANY($4)
```
bajtos, line 97: It's important to have an index in place to make this WHERE condition fast.

Please sync with @NikolasHaimerl about the UNIQUE index/constraint we want to add for active deals - see https://github.com/CheckerNetwork/deal-observer/pull/31/files#r1933931452

Depending on which columns uniquely identify a deal, you may need to add more of them to this WHERE condition.

backend/lib/spark-api-deal-submitter.js
```diff
@@ -0,0 +1,127 @@
...
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${dealIngestionAccessToken}`
+    },
+    body: JSON.stringify(unsubmittedDeals)
+  })
+
+  if (!response.ok) {
+    throw new Error(`Failed to submit deals: ${response.statusText}`)
```
bajtos, line 125: Please include the status code in the response. I believe the HTTP spec allows the server to return arbitrary status text; it's only the status code number that's standardised.

In my experience, it's difficult to troubleshoot failed requests when we have only the status, because the error details are usually found in the response body.

IMO, it would be best to extract this line into a helper function shared in all places in the backend component.

https://github.com/CheckerNetwork/deal-observer/blob/1c3471def968f3b55302dbc263b1e06c91dec123/backend/lib/rpc-service/service.js#L28

db/migrations/004.do.active-deals-submitted-at.sql
bajtos: We usually create a single migration script per pull request, unless there is a good reason for splitting.

I vaguely remember that postgrator runs each script in one transaction, and certain alterations cannot be executed together in one transactions, so we had to move them to a new migration script. If that was the case, you would know, because `npm run migrate` would fail.

backend/lib/spark-api-deal-submitter.js
bajtos, line 21: > In the case of an error during submission we won't retry to submit deals again. Rather submission loop will pick up unsubmitted deals next time it runs and will try to post them again.

If it's reasonably easy, then please consider adding a unit-test to verify this behaviour.

backend/lib/spark-api-deal-submitter.js
```diff
@@ -0,0 +1,127 @@
...
+  await pgPool.query(`
+    UPDATE active_deals
+    SET submitted_at = NOW()
+    WHERE miner_id = ANY($1) AND client_id = ANY($2) AND piece_cid = ANY($3) AND piece_size = ANY($4)
+  `, [
+    eligibleDeals.map(deal => deal.miner_id),
+    eligibleDeals.map(deal => deal.client_id),
+    eligibleDeals.map(deal => deal.piece_cid),
+    eligibleDeals.map(deal => deal.piece_size)
+  ])
```
bajtos, line 103: @pyropy I want to double-check that this query will update only the deals where the combination of `(miner_id, client_id, piece_cid, piece_size)` matches an existing combination in `eligibleDeals`.

We need this:

```sql
WHERE
  (miner_id = 'f010' AND client_id = 'f020' AND piece_cid = 'baga1' AND piece_size = 1)
  OR 
  (miner_id = 'f050' AND client_id = 'f060' AND piece_cid = 'baga2' AND piece_size = 1)
```

We need to avoid the following:

```sql
WHERE
  (miner_id = 'f010' OR miner_id = 'f050')
  AND 
  (client_id = 'f020' OR client_id = 'f060')
  AND 
  (piece_cid = 'baga1' OR piece_cid = 'baga2')
  AND
  (piece_size = 1 OR piece_size = 1)
```

Please start with ensuring there is a unit test for this.

If my suspicion is correct, I think you will need the `UNNEST` operator or something similar.

bajtos (changes requested):

juliangruber:
> How is your implementation going to handle errors? In particular, if the HTTP request to spark-api fails.

For now we can simply wait until the next loop iteration happens

backend/bin/deal-observer-backend.js
juliangruber: I have removed the currying

backend/lib/spark-api-deal-submitter.js
pyropy: You're right! It indeed updates other rows. 

I have changed query to use `UNNEST` and have added better mock data to unit tests to cover this scenario.

backend/lib/spark-api-deal-submitter.js
juliangruber: resolving due to lack of time

PR #37, opened 2025-01-28
fix: add token authentication

NikolasHaimerl:
THis PR adds token authentication for the GLIF RPC API to the rpc fetch method. 
The GLIF token is stored in the Fly.io instance and we also need to add the authorization header to the rpcRequest function.


backend/lib/rpc-service/service.js
```diff
@@ -14,9 +14,13 @@ import { ClaimEvent, RawActorEvent, BlockEvent, RpcRespone } from './data-types.
  */
 export const rpcRequest = async (method, params) => {
   const reqBody = JSON.stringify({ method, params, id: 1, jsonrpc: '2.0' })
+  const headers = { 'content-type': 'application/json' }
+  if (rpcHeaders.Authorization) {
+    Object.assign(headers, rpcHeaders)
+  }
```
bajtos, line 20: This can be simplified as follows:

```suggestion
  const headers = { 
    ...rpcHeaders,
    'content-type': 'application/json'
  }
```

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

Also note that header names are case-insensitive (at least in JavaScript), `rpcHeaders.authorization` and `rpcHeaders.Authorization` works the same way, so it's best not to rely on a specific property name here, and use the spread syntax instead.

bajtos (changes requested):

juliangruber (approved):

bajtos (approved):
:shipit:

PR #39, opened 2025-01-28
chore: move database access

NikolasHaimerl:
This PR extracts any database related function into the database module of the deal-observer repository. 
This change is done to keep the code organized and not mix domains in between the different code parts. 
This PR is extracted from [here](https://github.com/filecoin-station/deal-observer/pull/31).


db/lib/database-access.js
```diff
@@ -0,0 +1,101 @@
+/** @import { BlockEvent } from '../../backend/lib/rpc-service/data-types.js' */
+/** @import { Static } from '@sinclair/typebox' */
+/** @import {Queryable} from '@filecoin-station/deal-observer-db' */
+import { ActiveDealDbEntry } from '@filecoin-station/deal-observer-db/lib/types.js'
```
bajtos, line 4: This creates a circular reference between `db` and `backend`. Circular references are almost always a bad thing.

Please move the typedefs (model definitions) to `db/lib` too.

db/lib/database-access.js
```diff
@@ -0,0 +1,101 @@
...
+import { Value } from '@sinclair/typebox/value'
+
+/**
+ * @param {Queryable} pgPool
+ * @returns {Promise<ActiveDealDbEntry | null>}
+ */
+export async function fetchDealWithHighestActivatedEpoch (pgPool) {
```
bajtos, line 11: What do you think about using a Repository design pattern, where we have a Repository interface providing the data access methods, and a Repository implementation that holds the reference to the pgPool instance?

See what I did in piece-indexer:

https://github.com/filecoin-station/piece-indexer/blob/main/repository/lib/redis-repository.js

There is nothing wrong with your functional-programming approach here, it's a valid direction too.

db/lib/database-access.js
```diff
@@ -0,0 +1,101 @@
...
+  const result = await loadDeals(pgPool, query)
+  return result.length > 0 ? result[0] : null
+}
+
+/**
+   * @param {Static<typeof BlockEvent>[]} activeDeals
+   * @param {Queryable} pgPool
+   * @returns {Promise<void>}
+   * */
+export async function storeActiveDeals (activeDeals, pgPool) {
```
bajtos, line 22: Let's modify all data-access functions to accept `pgPool` as the first argument please, so that we have a consistent pattern used by all functions, and it's possible to use optional arguments.

Right now, there is inconsistency - `storeActiveDeals (activeDeals, pgPool)` vs `loadDeals (pgPool, query)`.

An illustration of an optional parameter:

```js
async function loadDeals (pgPool, query = 'SELECT * FROM active_deals') {
  const result = (await pgPool.query(query)).rows.map(deal => {
    return Value.Parse(ActiveDealDbEntry, deal)
  }
  )
  return result
}

console.log(await loadDeals(pgPool))
```

db/lib/database-access.js
```diff
@@ -0,0 +1,101 @@
+/** @import { BlockEvent } from '../../backend/lib/rpc-service/data-types.js' */
+/** @import { Static } from '@sinclair/typebox' */
+/** @import {Queryable} from '@filecoin-station/deal-observer-db' */
+import { ActiveDealDbEntry } from '@filecoin-station/deal-observer-db/lib/types.js'
+import { Value } from '@sinclair/typebox/value'
+
+/**
+ * @param {Queryable} pgPool
+ * @returns {Promise<ActiveDealDbEntry | null>}
```
bajtos, line 9: Please apply the changes from https://github.com/filecoin-station/deal-observer/commit/b6cf42ec3802b4fb8171b08530581a510a783a33 to fix the typings.

backend/test/deal-observer.test.js
```diff
@@ -1,10 +1,10 @@
 import assert from 'node:assert'
 import { after, before, beforeEach, describe, it } from 'node:test'
 import { createPgPool, migrateWithPgClient } from '@filecoin-station/deal-observer-db'
-import { fetchDealWithHighestActivatedEpoch, storeActiveDeals } from '../lib/deal-observer.js'
 import { ActiveDealDbEntry } from '@filecoin-station/deal-observer-db/lib/types.js'
 import { Value } from '@sinclair/typebox/value'
 import { BlockEvent } from '../lib/rpc-service/data-types.js'
+import { fetchDealWithHighestActivatedEpoch, storeActiveDeals } from '@filecoin-station/deal-observer-db/lib/database-access.js'
```
bajtos, line 7: Is there any particular reason why you are importing data-access functions from a specific file inside the other module instead of exposing data-access functions in the public API of `@filecoin-station/deal-observer-db`?

I vaguely remember importing from a specific file local to a 3rd-party-package file can break under certain circumstances, but I don't remember the details. 

Personally, I prefer this:

```js
import { 
  createPgPool, 
  fetchDealWithHighestActivatedEpoch,
  storeActiveDeals, 
  migrateWithPgClient,
} from '@filecoin-station/deal-observer-db'
```

@juliangruber WDYT? Do you share my concern about importing from files in module's `lib` folder?

bajtos (changes requested):
I love the direction 👏🏻 

Let's discuss the implementation.

db/lib/database-access.js
NikolasHaimerl: It does [live](https://github.com/filecoin-station/deal-observer/blob/main/db/lib/types.js#L14) in `db/lib`.

db/lib/database-access.js
NikolasHaimerl: Sounds good to me, I created an [issue](https://github.com/filecoin-station/deal-observer/issues/44) for this change.

backend/bin/deal-observer-backend.js
```diff
@@ -5,8 +5,9 @@ import slug from 'slug'
 import '../lib/instrument.js'
 import { createInflux } from '../lib/telemetry.js'
 import { getChainHead, rpcRequest } from '../lib/rpc-service/service.js'
-import { fetchDealWithHighestActivatedEpoch, observeBuiltinActorEvents } from '../lib/deal-observer.js'
+import { observeBuiltinActorEvents } from '../lib/deal-observer.js'
 import assert from 'node:assert'
+import { fetchDealWithHighestActivatedEpoch } from '@filecoin-station/deal-observer-db/lib/database-access.js'
```
juliangruber, line 10: I'm not sure I agree with the choice of moving all queries to the db module. If you extrapolate this pattern, you'll have a repo with 10 workspaces, but all the queries are bundled in a single workspace. I don't think this scales well, and think that keeping queries close to where they are being used (here inside `backend`) works better

juliangruber (changes requested):

backend/bin/deal-observer-backend.js
juliangruber: Say you have 10 workspaces, each needing sql queries. Then all of their queries will be inside the deal observer lib. I think this is the wrong abstraction. It's like putting the libraries of all workspaces into a global lib/ folder, or the READMEs of all workspaces into a global docs/ folder. Or would you prefer that pattern?

To avoid a potential source of confusion: Here I mean scaling in terms of scaling a code base, not performance scaling.

db/lib/database-access.js
bajtos: Oh, I see. In that case use a relative path in the import statement.

```suggestion
import { ActiveDealDbEntry } from './types.js'
```

backend/bin/deal-observer-backend.js
bajtos: I see your point, @juliangruber. 

Let's say we have a very specific query used in one place, e.g. "list me all miners and for each miner, count the number of deals per each client" called from a REST API endpoint. Under this proposal, the query would go to `db`, making the REST API endpoint implementation slightly more to reason about, because the SQL query is far away.

Now add a dozen or two of such endpoints, and we have a dozen or two of very specific queries inside `db`. As a result, when reading the code in `db`, it's not clear what these queries are good for, who is calling them and why.

It also creates a temptation to try to generalise these queries so that one data-access function can be reused from two or more places that need something similar but slightly different, which is often a slippery slope leading to overly complex code full of conditional branching, code that's difficult to maintain and often hides subtle bugs.

From that perspective, I agree it's better to keep SQL queries close to the code executing them.

I also see the value in what @NikolasHaimerl is trying to achieve - avoid duplicate implementation of common data-access patterns like "write an array of active deals to DB", "load (all) active deals from the DB", and so on.

Maybe there is a middle ground where we move generic read/write data-access functions inside `db` for easy reuse but keep more specialised queries close to the code that executes them.

Thoughts?

backend/lib/deal-observer.js
```diff
@@ -1,11 +1,10 @@
 /** @import {Queryable} from '@filecoin-station/deal-observer-db' */
-/** @import { BlockEvent } from './rpc-service/data-types.js' */
 /** @import { Static } from '@sinclair/typebox' */
+/** @import { ActiveDealDbEntry } from '@filecoin-station/deal-observer-db/lib/types.js' */
 
 import assert from 'node:assert'
 import { getActorEvents, getActorEventsFilter } from './rpc-service/service.js'
-import { ActiveDealDbEntry } from '@filecoin-station/deal-observer-db/lib/types.js'
-import { Value } from '@sinclair/typebox/value'
+import { loadDeals, storeActiveDeals } from '@filecoin-station/deal-observer-db'
```
juliangruber, line 7: the function `loadDeals` is only used in one place (here), why did you decide to move it the central place? I thought we would only do that for queries that are used in multiple workspaces

juliangruber: Similar for `storeActiveDeals`: It is only used in this lib, and this lib's tests. To reuse it from tests, rather let the test consume `sotreActiveDeals` exported by this lib

juliangruber (changes requested):

backend/lib/deal-observer.js
NikolasHaimerl: I opened the PR for piece indexer where I did quite a bit of refactoring. One change request was to outsource these refactorings in their separate PR. This is one of these PRs.

juliangruber: I understand, and that was the right thing to do. It is easier to discuss refactorings if they are decoupled from other work. Looking at the proposed changes now, I don't think these refactorings improve the code base, so I propose to close the PR. We agreed to DRY up queries if they are used in multiple places, and none of the one refactored here are. I'm sorry for your time spent, but sometimes that's how it goes with proposals - not all can make it.

PR #40, opened 2025-01-28
fix:  lastInsertedDeal wrong field

NikolasHaimerl:
This PR fixes the issue of calling the wrong field name for the last inserted deal. 
A deal contains the field `activated_at_epoch` while the BlockEvent contains the field `height`.


bajtos (approved):
Nice catch!

I added b6cf42e to fix the typings in such way that TypeScript would be able to detect this bug.

I highly recommend you to fix all other places using TypeBox `T.Object` values to use `Static<typeof T>` instead. But that's out of the scope of this pull request.

NikolasHaimerl:
> Nice catch!
> 
> I added [b6cf42e](https://github.com/filecoin-station/deal-observer/commit/b6cf42ec3802b4fb8171b08530581a510a783a33) to fix the typings in such way that TypeScript would be able to detect this bug.
> 
> I highly recommend you to fix all other places using TypeBox `T.Object` values to use `Static<typeof T>` instead. But that's out of the scope of this pull request.

I added an [issue](https://github.com/filecoin-station/deal-observer/issues/43) for this.

PR #41, opened 2025-01-28
fix: delete db active_deals table after test

NikolasHaimerl:
We need to delete the table `active_deals` after all the tests are done to avoid the tests leaving the database altered after the tests are run.

### DIFF
--- a/backend/test/deal-observer.test.js
+++ b/backend/test/deal-observer.test.js
@@ -14,6 +14,7 @@ describe('deal-observer-backend', () => {
   })
 
   after(async () => {
+    await pgPool.query('DELETE FROM active_deals')
     await pgPool.end()
   })
 


bajtos (changes requested):
No no no, we want to keep the data created by tests so that we can inspect the database contents when a test fails.

You should instead reset the database to an initial/empty state before each test to prevent tests from interfering with each other.

What problem are you trying to solve?

NikolasHaimerl:
> 

No no no, we want to keep the data created by tests so that we can inspect the database contents when a test fails.

You should instead reset the database to an initial/empty state before each test to prevent tests from interfering with each other.

What problem are you trying to solve?


Is there a scenario where inspecting the test database tells us more than the test messages already do? 
I added this because it left my local database instance contaminated any time I ran `npm test`.

juliangruber:
> > No no no, we want to keep the data created by tests so that we can inspect the database contents when a test fails.You should instead reset the database to an initial/empty state before each test to prevent tests from interfering with each other. What problem are you trying to solve?
> 
> Is there a scenario where inspecting the test database tells us more than the test messages already do? I added this because it left my local database instance contaminated any time I ran `npm test`.

I personally never run the services in development, only the test suite. I therefore don't care what's in my local db. Ideally a test takes are of resetting any state, but if it wouldn't, I'm using this script for resetting my dbs:

```console
alias smdbs='docker rm -f meridian-db && docker run --name meridian-db -e POSTGRES_HOST_AUTH_METHOD=trust -e POSTGRES_USER=$USER -e POSTGRES_DB=$USER -p 5432:5432 -d postgres && sleep 1; psql postgres://localhost:5432/ -c "CREATE DATABASE spark_evaluate" && psql postgres://localhost:5432/ -c "CREATE DATABASE spark_stats" && psql postgres://localhost:5432/ -c "CREATE DATABASE spark" && psql postgres://localhost:5432/ -c "CREATE DATABASE spark_rewards" && psql postgres://localhost:5432/ -c "CREATE DATABASE spark_deal_observer"'
```

> I added this because it left my local database instance contaminated any time I ran npm test.

For my understanding, how are your using your local database after running tests?

If you do care about keeping state in your DB, what do you think about configuring your environment so that tests talk to a different database?

NikolasHaimerl:
> > > No no no, we want to keep the data created by tests so that we can inspect the database contents when a test fails.You should instead reset the database to an initial/empty state before each test to prevent tests from interfering with each other. What problem are you trying to solve?
> > 
> > 
> > Is there a scenario where inspecting the test database tells us more than the test messages already do? I added this because it left my local database instance contaminated any time I ran `npm test`.
> 
> I personally never run the services in development, only the test suite. I therefore don't care what's in my local db. Ideally a test takes are of resetting any state, but if it wouldn't, I'm using this script for resetting my dbs:
> 
> ```
> alias smdbs='docker rm -f meridian-db && docker run --name meridian-db -e POSTGRES_HOST_AUTH_METHOD=trust -e POSTGRES_USER=$USER -e POSTGRES_DB=$USER -p 5432:5432 -d postgres && sleep 1; psql postgres://localhost:5432/ -c "CREATE DATABASE spark_evaluate" && psql postgres://localhost:5432/ -c "CREATE DATABASE spark_stats" && psql postgres://localhost:5432/ -c "CREATE DATABASE spark" && psql postgres://localhost:5432/ -c "CREATE DATABASE spark_rewards" && psql postgres://localhost:5432/ -c "CREATE DATABASE spark_deal_observer"'
> ```
> 
> > I added this because it left my local database instance contaminated any time I ran npm test.
> 
> For my understanding, how are your using your local database after running tests?
> 
> If you do care about keeping state in your DB, what do you think about configuring your environment so that tests talk to a different database?

I am a bit surprised that our testing setup would allow for an altered database state compared to before running the tests after it has completed all tests. I would have assumed that if the database was fresh and not records have been inserted that this would also be the state in which it is after the tests. 

I could of course setup multiple databases locally but it seems odd to me that tests alter database states. It might make sense to not run the tests against a dockerized postgres instance but spin up an in [memory](https://www.sqlite.org/inmemorydb.html) sql instance which only lives for as long as the test runs and only exists in memory.

juliangruber:
> I am a bit surprised that our testing setup would allow for an altered database state compared to before running the tests after it has completed all tests. I would have assumed that if the database was fresh and not records have been inserted that this would also be the state in which it is after the tests.

Could you please help us understand why this is a surprise to you? As mentioned before, @bajtos and me don't have such an expectation. Is there a source you can quote, or could you explain why that is better?

> I could of course setup multiple databases locally but it seems odd to me that tests alter database states. It might make sense to not run the tests against a dockerized postgres instance but spin up an in [memory](https://www.sqlite.org/inmemorydb.html) sql instance which only lives for as long as the test runs and only exists in memory.

Again why is it odd for tests to alter database states? I'm very happy to be wrong here, I just don't know better 😅

NikolasHaimerl:
> > I am a bit surprised that our testing setup would allow for an altered database state compared to before running the tests after it has completed all tests. I would have assumed that if the database was fresh and not records have been inserted that this would also be the state in which it is after the tests.
> 
> Could you please help us understand why this is a surprise to you? As mentioned before, @bajtos and me don't have such an expectation. Is there a source you can quote, or could you explain why that is better?
> 
> > I could of course setup multiple databases locally but it seems odd to me that tests alter database states. It might make sense to not run the tests against a dockerized postgres instance but spin up an in [memory](https://www.sqlite.org/inmemorydb.html) sql instance which only lives for as long as the test runs and only exists in memory.
> 
> Again why is it odd for tests to alter database states? I'm very happy to be wrong here, I just don't know better 😅

It is just that I do not expect that running tests changes the state of my machine. They should be agnostic to the underlying infrastructure IMO.
I do not think this discussion brings us closer to shipping the DDO deal though so I'll close this PR.

juliangruber:
Let's continue this discussion in another place if you wish to change this behavior 🙏

bajtos:
I apologise for the tone of my initial comment.

In the past, I experienced subtle bugs where I need to examine the state of the database after a test failed to find the cause.  However, after I honestly reflected on this, I don't remember the last time I ran into an issue like that. Maybe we don't have to optimise for "what if" scenarios like this. In that light, I no longer object to deleting the DB after the test suite finishes.

It would be nice if there was an easy way how to disable that deleting behaviour, e.g. by setting an ENV variable, or commenting out a single line of code.

On the other hand, I also think the test should be free to do whatever they want with the database, and it's better to have one database for running the service locally and a different database used by tests.

PR #42, opened 2025-01-28
chore: extract deal observer loop

NikolasHaimerl:
This PR introduces the following changes:

1. Extract the logic of the deal observer loop into its own file 
2. Add a test for the deal observer loop
3. Extract the mocked rpc endpoint into its own file for easier reuse.


backend/test/deal-observer.test.js
```diff
@@ -74,3 +76,47 @@ describe('deal-observer-backend', () => {
...
+        10_000,
+        undefined
+      )
+    }
+    // Timeout to kill the worker after 20 seconds
+    setTimeout(() => {
+      console.log('Condition not met, killing the worker.')
+      process.exit(1)
+    }, 20000)
+    worker()
```
bajtos, line 112: I am afraid this is not how Node.js works. When you call `worker()`, it starts in the same process. When you call `process.exit()`, you kill the entire test suite.

I have two suggestions for you to consider:

(1)
Move the `while(true)` loop outside of the function that implements the things we want to execute repeatedly.

This is the design we have in spark-stats, and it's the design I put in place in my initial pull request setting up the infrastructure in this repo.

https://github.com/filecoin-station/spark-stats/blob/2d20ebc6a524161d6c53e41c604e40ff87d326ee/observer/bin/spark-observer.js#L36-L75

https://github.com/filecoin-station/deal-observer/blob/e12a7725d62d7f5669d8be8a5cf846e93dce0a97/backend/bin/deal-observer-backend.js#L48-L54

This is my preferred approach because it's consistent with what we do in other repositories. There is no need to reinvent the wheel here, IMO.

(2)
Use [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) to cancel the infinite loop when running the loop function from tests.

Illustration:

```js
export const dealObserverLoop = async (/* ... */, signal) => {
  while (!signal?.aborted) {
    // ...
  }
}

// in tests
await dealObserverLoop(/* ... */, AbortSignal.timeout(20_000))
```

Having said that, waiting for 20 seconds will make the test suite very slow to run, especially if there is more than one test using this pattern.

A better solution is to frequently check whether the function under test has progressed enough so that you can verify the behaviour, or even better, implement the function to signal when that happens, e.g. using an event emitter or by calling a callback function provided in parameters.

As you can see, it is complex to get this right, that's why I prefer the first option above. 

Also, because the `loop` function is only periodically invoking the provided callback, it is much easier to test it if we wanted to, because the tests can provide a specific function optimised for the test case.

```js
// a mock-up to illustrate what I mean - this test finishes in <200ms
const ticks = []
const fn = () => ticks.push(Date.now())
await loop("test fn", fn, 100, AbortController.timeout(150))
// assert that `ticks` contain 2 entries that are approx. 100ms apart
```

backend/lib/loops.js
bajtos, line 1: Can we find a more descriptive file name please?

How about using one file per loop, and encoding the loop name in the file name?

E.g. `lib/deal-observer-loop.js`

bajtos (changes requested):
Great idea to move the loop functions from `bin/` to `lib/`!

backend/test/deal-observer.test.js
NikolasHaimerl: > I am afraid this is not how Node.js works. When you call `worker()`, it starts in the same process. When you call `process.exit()`, you kill the entire test suite.
> 
> I have two suggestions for you to consider:
> 
> (1) Move the `while(true)` loop outside of the function that implements the things we want to execute repeatedly.
> 
> This is the design we have in spark-stats, and it's the design I put in place in my initial pull request setting up the infrastructure in this repo.
> 
> https://github.com/filecoin-station/spark-stats/blob/2d20ebc6a524161d6c53e41c604e40ff87d326ee/observer/bin/spark-observer.js#L36-L75
> 
> https://github.com/filecoin-station/deal-observer/blob/e12a7725d62d7f5669d8be8a5cf846e93dce0a97/backend/bin/deal-observer-backend.js#L48-L54
> 
> This is my preferred approach because it's consistent with what we do in other repositories. There is no need to reinvent the wheel here, IMO.
> 
> (2) Use [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) to cancel the infinite loop when running the loop function from tests.
> 
> Illustration:
> 
> ```js
> export const dealObserverLoop = async (/* ... */, signal) => {
>   while (!signal?.aborted) {
>     // ...
>   }
> }
> 
> // in tests
> await dealObserverLoop(/* ... */, AbortSignal.timeout(20_000))
> ```
> 
> Having said that, waiting for 20 seconds will make the test suite very slow to run, especially if there is more than one test using this pattern.
> 
> A better solution is to frequently check whether the function under test has progressed enough so that you can verify the behaviour, or even better, implement the function to signal when that happens, e.g. using an event emitter or by calling a callback function provided in parameters.
> 
> As you can see, it is complex to get this right, that's why I prefer the first option above.
> 
> Also, because the `loop` function is only periodically invoking the provided callback, it is much easier to test it if we wanted to, because the tests can provide a specific function optimised for the test case.
> 
> ```js
> // a mock-up to illustrate what I mean - this test finishes in <200ms
> const ticks = []
> const fn = () => ticks.push(Date.now())
> await loop("test fn", fn, 100, AbortController.timeout(150))
> // assert that `ticks` contain 2 entries that are approx. 100ms apart
> ```

Thanks for the tip. I modified the testaccordingly.

backend/bin/deal-observer-backend.js
```diff
@@ -19,42 +17,18 @@ const finalityEpochs = 940
...
   rpcRequest,
-  pgPool
+  pgPool,
+  recordTelemetry,
+  Sentry,
+  maxPastEpochs,
+  finalityEpochs,
+  LOOP_INTERVAL,
+  INFLUXDB_TOKEN,
+  signal
```
juliangruber, line 33: This method has a lot of arguments, working with positional arguments is hard in this case. Imagine using an editor that doesn't infer types, or if we later on change the arguments list.

What do you think about changing this to an options object?

backend/test/utils.js
```diff
@@ -0,0 +1,14 @@
+import { parse } from '@ipld/dag-json'
+import { chainHeadTestData } from './test_data/chainHead.js'
+import { rawActorEventTestData } from './test_data/rawActorEvent.js'
+
+export const makeRpcRequest = async (method, params) => {
```
juliangruber, line 5: I'm not sure this will scale well. @bajtos what do you think about this helper? One test might need a different test data set being returned than another. My suspicion is that it will scale better to define this `makeRpcRequest` function where the tests are also defined (and potentially doing it multiple times).

juliangruber (changes requested):

backend/bin/deal-observer-backend.js
bajtos: +1 to use an options object (named parameters).

backend/test/deal-observer.test.js
```diff
@@ -74,3 +76,40 @@ describe('deal-observer-backend', () => {
     assert.deepStrictEqual(expected, actual)
   })
 })
+
+describe('deal-observer-backend binary', () => {
```
bajtos, line 80: Since this test is no longer testing the bin script, but only the loop, let's rename it accordingly:

```suggestion
describe('dealObserverLoop', () => {
```

I find it difficult to navigate in very large files, I prefer to have more smaller files instead. Since these tests are testing code from `backend/lib/deal-observer-loop.js`, I would like the test code to be in `backend/test/deal-observer-loop.test.js`. 

It's a convention we already use in other repositories. It makes it easy to find tests because we can navigate using file names instead of scanning a long test file containing tests for many different components to find the right `describe` block.

backend/test/deal-observer.test.js
```diff
@@ -74,3 +76,40 @@ describe('deal-observer-backend', () => {
...
+  before(async () => {
+    pgPool = await createPgPool()
+    await migrateWithPgClient(pgPool)
+  })
+  beforeEach(async () => {
+    await pgPool.query('DELETE FROM active_deals')
+  })
+  after(async () => {
+    await pgPool.query('DELETE FROM active_deals')
+  })
```
bajtos, line 91: Would you mind not cleaning the DB after tests in this pull request and waiting until we resolve the discussion in https://github.com/filecoin-station/deal-observer/pull/41?

backend/test/deal-observer.test.js
```diff
@@ -74,3 +76,40 @@ describe('deal-observer-backend', () => {
...
+  it('then deal observer loop fetches new active deals and stores them in storage', async (t) => {
+    const controller = new AbortController()
+    const { signal } = controller
+    dealObserverLoop(
```
bajtos, line 95: This leaks a promise that's not awaited. When `dealObserverLoop` throws an unhandled error (promise rejection), this error won't be caught by your test.

In this particular case, it should be okay, because unhandled rejections are reported to console and your test waits until dealObserverLoop created all deals, at which point it's unlikely that dealObserverLoop has not finished running yet.

However, I prefer to make our code easy to reason about without the need to understand intricate details like how unhandled rejections are treated by the runtime and the test runner.

Here is one way how to implement this in a cleaner way:

```js
const failOnTimeout = async () => {
  await setTimeout(1000)
  throw new Error('test timed out')
}

const waitForDealCount = async (targetCount) => {
  do {
    rows = (await pgPool.query('SELECT * FROM active_deals')).rows
  } while (rows.length !== targetCount)
}

await Promise.all([
  failOnTimeout(),
  waitForDealCount(360),
  dealObserverLoop(/*...*/)
)
```
  
(My code uses the `Promise` version of `setTimeout`, see https://nodejs.org/api/timers.html#timerspromisessettimeoutdelay-value-options)

backend/test/deal-observer.test.js
```diff
@@ -74,3 +76,40 @@ describe('deal-observer-backend', () => {
...
+      1,
+      undefined,
+      signal
+    )
+    // Timeout to kill the worker after 1 second
+    let rows
+    setTimeout(() => { if (!signal.aborted) { throw new Error(`Test timed out. Rows inserted ${rows.length}`) } }, 1000)
+    do {
+      rows = (await pgPool.query('SELECT * FROM active_deals')).rows
+    } while (rows.length !== 360)
```
bajtos, line 112: Nitpick - feel free to ignore.

Fetching data for all rows is not a very efficient way to count the number of rows in a table. It does not matter in this test, since you are fetching up to 360 rows, but it could become a problem if used elsewhere.

A more efficient solution is to use the SQL aggregate function `COUNT()`.

```js
while (true) {
  const { rows } = await pgPool.query('SELECT COUNT(*) FROM active_deals')
  if (rows[0].count === 360) break
}
```

backend/test/utils.js
bajtos: I share your concerns and agree it's best to define the stub `makerRpcRequest` function in every test.

We can implement a builder helper to make it easier to implement such stubs.

```js
// usage in tests
const makerRpcRequest = buildMakeRpcRequest({
  'Filecoin.ChainHead': chainHeadTestData
})

// implementation
export const buildMakeRpcRequest = (methodsToResults) => {
  return async (method, _params) => {
    if (methodsToResults[method]) return methodsToResults[method]
    throw new Error(`Unsupported RPC API method: "${method}"`)
  }
}

File: backend/test/utils.js [hunk 8]
```diff
@@ -0,0 +1,14 @@
...
+  switch (method) {
+    case 'Filecoin.ChainHead':
+      return parse(JSON.stringify(chainHeadTestData))
+    case 'Filecoin.GetActorEventsRaw':
+      return parse(JSON.stringify(rawActorEventTestData)).filter(e => e.height >= params[0].fromHeight && e.height <= params[0].toHeight)
+    default:
+      console.error('Unknown method')
```
bajtos, line 12: This does not feel right to me. The callers of this stub function will receive `undefined` as the result, I am concerned we will often forget to handle that edge case.

I propose throwing an error instead.

```js
throw new Error(`Unsupported RPC API method: "${method}"`)
```

bajtos (changes requested):

backend/bin/deal-observer-backend.js
NikolasHaimerl: @bajtos  isn't this in contrast to what you mentioned in an earlier [PR](https://github.com/CheckerNetwork/deal-observer/pull/31#discussion_r1932185558)?

juliangruber: In the earlier PR @bajtos also suggested to use an options object / param object / named parameters

NikolasHaimerl: I see, I must have misunderstood then.

bajtos: > @bajtos isn't this in contrast to what you mentioned in an earlier [PR](https://github.com/CheckerNetwork/deal-observer/pull/31#discussion_r1932185558)?

It's great that you asked. As Julian pointed out, I meant the same thing in that other comment.

JavaScript language does not provide syntax for named parameters. We implement named parameters by passing around an object where each property corresponds to one parameter.

```js
// definition
function doSomeStuff({ pgClient, signal }) {
  // ...
}

// example usage
const pgClient = /*..*/

doSomeStuff({
  pgClient,
  signal: AbortSignal.timeout(10_000)
})
```

bajtos: The TypeScript typings are slightly more complex - you need to describe the single positional parameter accepting an object, and then describe the properties of that object.

```js
/**
 * @param {object} params
 * @param {pg.Client} params.pgClient
 * @param {AbortSignal} signal
 */
function doSomeStuff({ pgClient, signal }) {
  // ...
}
```

I personally tend to call the object argument as `args`, see e.g. here:


https://github.com/CheckerNetwork/piece-indexer/blob/765dbfe7705ce375f7b5a317f4912264e7a70ffa/indexer/lib/ipni-watcher.js#L11-L16

```js
/**
 * @param {object} args
 * @param {number} args.minSyncIntervalInMs
 * @param {AbortSignal} [args.signal]
 */
export async function * runIpniSync ({ minSyncIntervalInMs, signal }) {
  // ...
}
```

The syntax `[args.signal]` tells the type-checker that `signal` is an optional parameter.

NikolasHaimerl: I am not sure I quite understand. How does 
`(minSyncIntervalInMs, signal)` differ from `(minSyncIntervalInMs, signal)` for readability. Or do you mean to have a single argument `(args)` and then call `args.pgPool`?

backend/lib/deal-observer-loop.js
```diff
@@ -0,0 +1,50 @@
...
+ * @param {Queryable} pgPool
+ * @param {(name: string, fn: (p: Point) => void) => void} recordTelemetry
+ * @param {import("@sentry/node")} Sentry
+ * @param {number} maxPastEpochs
+ * @param {number} finalityEpochs
+ * @param {number} loopInterval
+ * @param {string | undefined} influxToken
+ * @returns {Promise<void>}
+ * */
+export const dealObserverLoop = async (makeRpcRequest, pgPool, recordTelemetry, Sentry, maxPastEpochs, finalityEpochs, loopInterval, influxToken, signal) => {
```
pyropy, line 19: Do you think that extracting the loop logic to a separate function could be a good idea? I had something like this in mind:

```javascript
// lib/loop.js
const loop = async ({ signal, name, interval, fn, recordTelemetry, captureException }) => {
  const submitDeals = submitDealsToSparkApi(sparkApiBaseURL, dealIngestionAccessToken)
  while (!signal?.aborted) {
    const start = Date.now()
    try {
      await fn()
    } catch (e) {
      console.error(e)
      captureException(e)
    }
    const dt = Date.now() - start
    console.log(`Loop "${name}" took ${dt}ms`)
    recordTelemetry(`loop_${slug(name, '_')}`, point => {
      point.intField('interval_ms', name)
      point.intField('duration_ms', dt)
    })

    if (dt < interval) {
      await timers.setTimeout(interval - dt)
    }
  }
}

// lib/deal-observer-loop.js
const dealObserverLoop = (pgPool, rpcClient, opts) => async () => {
  // logic goes here
}

// lib/deal-submitter-loop.js
const dealSubmitLoop = (pgPool, opts) => async () => {
  // logic goes here
}

// bin/deal-observer-backend.js
await Promise.all([
  loop({
    name: 'dealObserver',
    interval: 1000,
    fn: dealObserverLoop(pgPool, rpcClient, opts),
    recordTelemetry: console.log,
    captureException: console.error,
  }),
  loop({
    name: 'dealSubmit',
    interval: 1000,
    fn: dealSubmitLoop(pgPool, opts),
    recordTelemetry: console.log,
    captureException: console.error,
  })
])
```
I feel that something like this could help us to keep code more DRY, while also keeping the loop function itself clean as we don't have to pass so many arguments around (sentry, influx, and other dependencies).

backend/test/utils.js
NikolasHaimerl: Isn"t this adding quite a bit of complexity for a simple problem?

juliangruber: Wouldn't it be less complex to inline this function (without) the helper, than creating a new utility (i.e. a new abstraction)

backend/lib/deal-observer-loop.js
juliangruber: As discussed, #33 can be landed without the loop refactor (can be followed up with)

NikolasHaimerl:
I close this PR because the PR's goal is outdated and there another issue which does something similar that already has a [PR](https://github.com/CheckerNetwork/spark-deal-observer/pull/81) that was merged.

PR #14, opened 2025-01-19
Refactor using codeowners

juliangruber:
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

PR #47, opened 2025-01-29
disable `CODEOWNERS` on pkg files, to unbreak automerge

juliangruber:
https://github.com/orgs/community/discussions/23064

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,3 @@
 * @bajtos @NikolasHaimerl
+*/package.json
+package-lock.json


bajtos (changes requested):
When we disable code owners on package files, then nobody will be assigned to review dependabot pull requests.

I guess that's fine for auto-approved pull request.

However, it also means we won't be notified about Dependabot pull requests requiring manual review.

Are we fine with that?

If not, then I think we need to revert `.github/dependabot.yml` changes made in #14.

WDYT?

juliangruber:
> However, it also means we won't be notified about Dependabot pull requests requiring manual review.

You will be if you're a watcher of the project, which we by default are. Does that work for you?

bajtos (approved):
> > However, it also means we won't be notified about Dependabot pull requests requiring manual review.
> 
> You will be if you're a watcher of the project, which we by default are. Does that work for you?

I guess watching the repositories for new pull requests works. 

The caveat is that, AFAIK, we are not becoming watchers of new repositories automatically when a new repository is created; I would be surprised if I remembered to watch all the repositories we have now. Also, when we rely on watching, we won't see Dependabot pull requests listed in https://github.com/pulls >> Review requests. Finally, it's hard to say who is responsible for reviewing Dependabot pull requests in which repo, because you cannot tell if I am watching that repo.

So, I think depending on watchers is not ideal. But this process is easy to tweak, so let's start somewhere and iteratively improve as needed.

bajtos:
I just checked and I am actually not watching this repository ([deal-observer](https://github.com/CheckerNetwork/deal-observer)). I'll start watching it.

juliangruber:
Screenshot 2025-01-29 at 13 26 35

There's a setting for it, which is enabled by default (see https://docs.github.com/en/account-and-profile/managing-subscriptions-and-notifications-on-github/setting-up-notifications/configuring-notifications#automatic-watching)

bajtos:
FWIW, it was not enabled in my notification settings. Maybe it's because I created my GH account long before this setting was introduced.

PR #54, opened 2025-01-30
fix: log request body

NikolasHaimerl:

1. If a RPC request fails we also want the request body to be added to the error message.

### DIFF
--- a/backend/lib/rpc-service/service.js
+++ b/backend/lib/rpc-service/service.js
@@ -36,7 +36,7 @@ export const rpcRequest = async (method, params) => {
       throw error
     }
   } catch (error) {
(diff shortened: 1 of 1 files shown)


backend/lib/rpc-service/service.js
```diff
@@ -36,7 +36,7 @@ export const rpcRequest = async (method, params) => {
       throw error
     }
   } catch (error) {
-    error.message = `Failed to make RPC request ${method}: ${error.message}`
+    error.message = `Failed to make RPC request ${method}.\n Request was: ${JSON.stringify(reqBody)}.\n Error message is: ${error.message}`
```
bajtos, line 39: A suggestion on how to shorten the message; feel free to ignore it.

```suggestion
    error.message = `Failed to make RPC request ${method}: ${error.message}`\nRequest was: ${JSON.stringify(reqBody)}.` 
```

bajtos (approved):

PR #56, opened 2025-01-30
fix: remove max past epochs

NikolasHaimerl:
The free tier of GLIF RPC api allows for a maximum of 100 requests per minute. If we start with 2000 blocks in the pasts we will get rate limiting errors in the beginning of the sync as we will make more calls than 100 per minute to catch up with the current head. 
To avoid this we can remove the max past epoch variable.

### DIFF
--- a/backend/bin/deal-observer-backend.js
+++ b/backend/bin/deal-observer-backend.js
@@ -15,6 +15,8 @@ const LOOP_INTERVAL = 10 * 1000
 // Filecoin will need some epochs to reach finality.
 // We do not want to fetch deals that are newer than the current chain head - 940 epochs.
 const finalityEpochs = 940
+// The free tier of the Glif RPC endpoint allows for 100 requests per minute.
+const maximumRequestsPerMinute = 100
 const pgPool = await createPgPool()
 
 const LOOP_NAME = 'Built-in actor events'
@@ -28,7 +30,9 @@ const dealObserverLoop = async (makeRpcRequest, pgPool) => {
       const currentFinalizedChainHead = currentChainHead.Height - finalityEpochs
       // If the storage is empty we start 2000 blocks into the past as that is the furthest we can go with the public glif rpc endpoints.
       const lastInsertedDeal = await fetchDealWithHighestActivatedEpoch(pgPool)
-      const lastEpochStored = lastInsertedDeal ? lastInsertedDeal.activated_at_epoch : currentFinalizedChainHead - 1
+      // If there are no deals in the database we start from the current chain head - and use the maximum number of requests we can perform per minute as a lower bound.
+      // This is to avoid hitting the rate limit of the glif rpc endpoint and introduces more recilience to database gaps.
+      const lastEpochStored = lastInsertedDeal ? lastInsertedDeal.activated_at_epoch : currentFinalizedChainHead - maximumRequestsPerMinute
       for (let epoch = lastEpochStored + 1; epoch <= currentFinalizedChainHead; epoch++) {
         await observeBuiltinActorEvents(epoch, pgPool, makeRpcRequest)
       }
--- a/backend/bin/deal-observer-backend.js
+++ b/backend/bin/deal-observer-backend.js
@@ -15,8 +15,6 @@ const LOOP_INTERVAL = 10 * 1000
 // Filecoin will need some epochs to reach finality.
 // We do not want to fetch deals that are newer than the current chain head - 940 epochs.
 const finalityEpochs = 940
-// The free tier of the Glif RPC endpoint allows for 100 requests per minute.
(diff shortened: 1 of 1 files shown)


juliangruber:
FYI we can also use a paid tier token, if that helps a lot

NikolasHaimerl:
> FYI we can also use a paid tier token, if that helps a lot

I vaguely remember that in the last monthly planing, we mentioned that we have to move to a paid version anyway soon. Is that the case for the GLIF API too?

backend/bin/deal-observer-backend.js
```diff
@@ -31,7 +30,9 @@ const dealObserverLoop = async (makeRpcRequest, pgPool) => {
       const currentFinalizedChainHead = currentChainHead.Height - finalityEpochs
       // If the storage is empty we start 2000 blocks into the past as that is the furthest we can go with the public glif rpc endpoints.
       const lastInsertedDeal = await fetchDealWithHighestActivatedEpoch(pgPool)
-      const lastEpochStored = lastInsertedDeal ? lastInsertedDeal.activated_at_epoch : currentChainHead.Height - maxPastEpochs
+      // If there are no deals in the database we start from the current chain head - and use the maximum number of requests we can perform per minute as a lower bound.
+      // This is to avoid hitting the rate limit of the glif rpc endpoint and introduces more recilience to database gaps.
+      const lastEpochStored = lastInsertedDeal ? lastInsertedDeal.activated_at_epoch : currentFinalizedChainHead - maximumRequestsPerMinute
```
bajtos, line 35: I propose to simplify this and start the ingestion from the current chain head. As discussed in Slack, we need to find a more robust solution for handling blocks added while the service is down, it's a high-priority task for February, so we can afford to simplify now.

Proposed solution (remember to account for +1 in the loop initialiser below!):

```js
const lastEpochStored = lastInsertedDeal 
  ? lastInsertedDeal.activated_at_epoch 
  : currentFinalizedChainHead - 1
```

A longer reply: I am not confident `-100` is the right number to use. Let me explain why:


When I read `currentFinalizedChainHead - maximumRequestsPerMinute`, it sparks a bunch of questions.
- We are combining position (head block number) with rate (req/min). That feels to me like adding `1m + 2m/s` in physics, which is a signal I need to dig deeper to better understand.
- To combine position (head block number) with number of blocks, I think we would need a formula taking into account 1) number of requests needed to process one epoch 2) how often we process a new epoch 3) number of requests allowed per minute.

   ```
   lastEpochStored = currentFinalizedChainHead - (maximumRequestsPerMinute/requestsWe MakePerEpoch/epochsWeProcessEveryMinute)
   ```

- The problem is that the value of "requests per epoch" can change as the code evolves. One could assume it's just one request per epoch - "get `claim` events for epoch N". But that's going to change soon, as are going to map MinerIds to their PeerIDs in https://github.com/CheckerNetwork/deal-observer/pull/31

- The value "epochsWeProcessEveryMinute" can also change - on the initial run, we process events as quickly as we can, until we reach the chain head. Then we need to make one RPC API call to get the current chain head + fetch events if there is a new head after waiting for the delay inside the top-level loop.

As you can see now, calculating the correct number of blocks to look behind is a tricky problem.

My problem with the formula `currentFinalizedChainHead - maximumRequestsPerMinute` is that it hides the actual complexity from the reader and creates a false sense of understanding the problem space. While it could work as a quick fix now, it's likely to cause subtle bugs in the future.

That's why I prefer to find a simpler solution, like the one I suggested at the top.

bajtos (changes requested):

backend/bin/deal-observer-backend.js
NikolasHaimerl: I see your point. The reason for me to use the 100 was an approach of "We get 100 requests per Minute, might as well use them". Your explanation makes sense, however, this solution was never meant to act as a longer-term solution. It just gives us a 30 minute time window to avoid gaps in the database instead of 30 seconds for now. 
We need to come up with a permanent solution anyway, which also involves closing certain gaps in the database if they occur. 
I will revert the last change and set it back to `currentFinalizedChainHead - 1`

bajtos (approved):
:shipit:

backend/bin/deal-observer-backend.js
bajtos: Makes sense. Here is the situation where you solution may broke:

- We start 10 seconds after a new block was minted
- We run the initial run, consume our quota of 100 requests/minute, and finish in <20s
- A new block is minted and we want to process another block
- We cannot fetch the current chain head because we have already exhausted the quota

I suppose this is a recoverable error, the top-level loop will pause for a bit and either the quota is refreshed or we repeat the pause and try again. But it would produce error logs and potentially trigger false alarms.

PR #77, opened 2025-02-03
feat: check for missing payload

NikolasHaimerl:
This PR adds the following changes:

1. Add a cache for payloads that could not be retrieved 
2. Retrieve payloads up to 4 times with a time delay of 8 hours in between calls
3. Mark deals as unretrievable if all retries fail

See also this [issue](https://github.com/CheckerNetwork/spark-deal-observer/issues/30).


backend/bin/deal-observer-backend.js
```diff
@@ -128,12 +129,13 @@ const sparkApiSubmitDealsLoop = async (pgPool, { sparkApiBaseUrl, sparkApiToken,
 
 export const pieceIndexerLoop = async (makeRpcRequest, getDealPayloadCid, pgPool) => {
   const LOOP_NAME = 'Piece Indexer'
+  const payloadsCache = new NodeCache()
```
juliangruber, line 132:
```suggestion
  const payloadsCache = new Map()
```

It looks like we're not using any node-catch features, so we can use a plain map instead

backend/lib/piece-indexer.js
```diff
@@ -1,37 +1,74 @@
...
-    const payloadCid = await getDealPayloadCid(minerPeerId, deal.piece_cid)
-    if (payloadCid) {
-      deal.payload_cid = payloadCid
+    const cacheKey = JSON.stringify({ minerPeerId, pieceCid: deal.piece_cid })
+    if (cache.has(cacheKey)) {
+      await checkCacheForRetrievablePayloads(minerPeerId, deal, getDealPayloadCid, cache, now)
+    } else {
+      deal.payload_cid = await getDealPayloadCid(minerPeerId, deal.piece_cid)
+      if (!deal.payload_cid) {
+        cache.set(cacheKey, { retriesLeft: MAX_RETRIES, lastRetry: now })
```
juliangruber, line 32: A cache is the wrong abstraction here. We're not caching values (where you'd store a value that is the result of an operation, and then on successive calls read from the cache instead of repeating the operation), instead we're storing meta information about values (retries left and last retry timestamp).

Did you mean to cache values instead, or is the goal to simply keep track of retries, and abort early if a value has been attempted to fetch too often?

juliangruber (changes requested):

juliangruber:
https://github.com/CheckerNetwork/spark-deal-observer/pull/31#discussion_r1934240694 is capturing the goal of this PR, right?

bajtos (changes requested):
I am not a fan of adding in-memory cache for several reasons.

1. The cached values are lost on service restart, e.g. after we commit a change to the `main` branch, which deploys a new version to production.
2. A cache adds complexity to our setup.
3. IIUC your design, on every loop iteration, you fetch all deals missing PayloadCID from the database. That will eventually mean fetching tens of millions of records.
3. While the solution proposed here does improve the situation, it also leaves out a lot of further opportunities to save the number of "get deal payload CID" calls. This is because most deals not advertised to IPNI will remain unadvertised even after we perform max retries, simply because most SPs don't advertise deals to IPNI now. We want to keep the retry mechanism to account for delays in IPNI processing new deals, which is important, but it applies to a very small fraction of deals.
  So, after the service is restarted, and the in-memory cache is empty, we have to run the initial query for all deals, even those we checked not so long ago.

Let's keep the state related to retrying in the database, please, and narrow down the SQL query loading deals to fetch only deals that we either did not check yet or the last check was done before 8 hours (or more).

Remember to add an SQL index to speed up the query. We want to avoid an expensive scan of all rows in the table.

backend/lib/piece-indexer.js
NikolasHaimerl: The goal is to keep track of which combination of minderID and pieceCID we have already tried to fetch and when we should try again or give up. The reason why I did use a cache (or some data store in memory) is because I wanted to keep the database clean of any information related to the state of retrievable or non retrievable piece CIDs in between fetching them for the first time and eventually giving up on fetching them.

bajtos:
> 3. IIUC your design, on every loop iteration, you fetch all deals missing PayloadCID from the database. That will eventually mean fetching tens of millions of records.

I miscalculated the number of deals here. IIUC, we mark deals as not advertised after ~1 day. fil-deal-ingester is adding 40-60k daily these days, and around 33% of deals are advertised to IPNI, according to what I see in the Spark Internal Dashboard. That means around 77% * 40k = ~30k deals to recheck after the service (re)starts.

I apologise for exaggerating the number in my initial comment.

I still prefer to keep our service as stateless as possible. It's a principle we have been following in other services, too. Please correct me if I am wrong, @juliangruber.

backend/lib/piece-indexer.js
juliangruber: Can you please add more detail? Eg:

- will we give up forever, or eventually try again?
- why did you want to keep the database clean of this information?
- if you are concerned about the table growing indefinitely, are there ways to use the DB but still prevent that?

NikolasHaimerl:
> > 3. IIUC your design, on every loop iteration, you fetch all deals missing PayloadCID from the database. That will eventually mean fetching tens of millions of records.
> 
> I miscalculated the number of deals here. IIUC, this PR will mark deals as not advertised after ~1 day. fil-deal-ingester is adding 40-60k daily these days. Around 33% of deals are advertised to IPNI, according to what I see in the Spark Internal Dashboard. That means around 77% * 40k = ~30k deals to recheck after the service (re)starts.
> 
> I apologise for exaggerating the number in my initial comment.
> 
> I still prefer to keep our service as stateless as possible. It's a principle we have been following in other services, too. Please correct me if I am wrong, @juliangruber.

I created an [implementation plan](https://github.com/CheckerNetwork/spark-deal-observer/issues/30#issuecomment-2631602358) which makes the service stateless in exchange for an additional field in the database to keep track of when the last time was when we tried to fetch a payload. Please let me know what you think.

backend/lib/piece-indexer.js
NikolasHaimerl: The goal of this implementation is to account for the delay that can occur for SP providers when publishing to IPNI. This delay was estimated to be less than 3 days. After 3 days the piece indexer will try again to fetch those deals and if the retry is unsuccessful it will mark the payload as unretrievable and give up on trying to fetch that payload. I wanted to keep the database clean initially because it would mean an additional column. that needs storage space . I updated my [implementation plan](https://github.com/CheckerNetwork/spark-deal-observer/issues/30#issuecomment-2631602358) and there I store the last retry in the database.

db/migrations/009.do.add-payload-retrievability-column.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE active_deals
```
NikolasHaimerl, line 1: I am still working on the index optimization.

backend/lib/piece-indexer.js
juliangruber: > The reason why I did use a cache (or some data store in memory)

To clarify: A data store in memory isn't usually a cache, and a cache isn't necessarily a data store in memory. See

> We're not caching values (where you'd store a value that is the result of an operation, and then on successive calls read from the cache instead of repeating the operation), instead we're storing meta information about values (retries left and last retry timestamp).

If we used a cache here, we could cache the result of the payload lookup, so that we wouldn't look up again. The cache would contain the value of the actual payload, and not just information about the lookup.

backend/lib/piece-indexer.js
```diff
@@ -43,12 +47,14 @@ export async function fetchDealsWithNoPayloadCid (pgPool, maxDeals) {
 async function updatePayloadInActiveDeal (pgPool, deal) {
   const updateQuery = `
     UPDATE active_deals
-    SET payload_cid = $1
-    WHERE activated_at_epoch = $2 AND miner_id = $3 AND client_id = $4 AND piece_cid = $5 AND piece_size = $6 AND term_start_epoch = $7 AND term_min = $8 AND term_max = $9 AND sector_id = $10
+    SET payload_cid = $1, payload_unretrievable = $2, last_payload_retrieval = $3
```
pyropy, line 50: We can also avoid passing down and mapping `last_payload_retrieval` value by simply using `NOW()`

```suggestion
    SET payload_cid = $1, payload_unretrievable = $2, last_payload_retrieval = NOW()
```

db/migrations/009.do.add-payload-retrievability-column.sql
```diff
@@ -0,0 +1,3 @@
...
+ADD COLUMN payload_unretrievable BOOLEAN,
+ADD COLUMN last_payload_retrieval TIMESTAMP WITH TIME ZONE;
```
pyropy, line 3: Have you tried running this migration with type set to `TIMESTAMPZ`?

```suggestion
ADD COLUMN last_payload_retrieval TIMESTAMPZ;
```

backend/lib/piece-indexer.js
```diff
@@ -14,25 +16,27 @@ import { getMinerPeerId } from './rpc-service/service.js'
  * @param {number} maxDeals
  * @returns {Promise<void>}
  */
-export const indexPieces = async (makeRpcRequest, getDealPayloadCid, pgPool, maxDeals) => {
-  for (const deal of await fetchDealsWithNoPayloadCid(pgPool, maxDeals)) {
+export const indexPieces = async (makeRpcRequest, getDealPayloadCid, pgPool, maxDeals, now = Date.now()) => {
+  for (const deal of await fetchDealsWithNoPayloadCid(pgPool, maxDeals, new Date(now - THREE_DAYS_IN_MILLISECONDS))) {
```
pyropy, line 20: If we calculate this date in query itself we can avoid having to pass `now` argument ending with simpler function signature.

db/migrations/009.do.add-payload-retrievability-column.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE active_deals
+ADD COLUMN payload_unretrievable BOOLEAN,
```
pyropy, line 2: Small nitpick, but have you been thinking about using `payload_retrievable` for column name? Seems a bit simpler to me.

pyropy (commented):
Good job, I think we're almost there.

backend/lib/piece-indexer.js
```diff
@@ -80,13 +47,14 @@ export async function fetchDealsWithNoPayloadCid (pgPool, maxDeals) {
 async function updatePayloadInActiveDeal (pgPool, deal) {
   const updateQuery = `
     UPDATE active_deals
-    SET payload_cid = $1, payload_unretrievable = $2
-    WHERE activated_at_epoch = $3 AND miner_id = $4 AND client_id = $5 AND piece_cid = $6 AND piece_size = $7 AND term_start_epoch = $8 AND term_min = $9 AND term_max = $10 AND sector_id = $11
+    SET payload_cid = $1, payload_unretrievable = $2, last_payload_retrieval = $3
+    WHERE activated_at_epoch = $4 AND miner_id = $5 AND client_id = $6 AND piece_cid = $7 AND piece_size = $8 AND term_start_epoch = $9 AND term_min = $10 AND term_max = $11 AND sector_id = $12
```
bajtos, line 51: Would it make it easier to update individual active deal rows if we added an auto-generated primary key so that we can simplify the WHERE condition to `WHERE id = $4` instead of comparing values in 9 column?

backend/lib/piece-indexer.js
bajtos: I think the goal is to make this function easier to test - by injecting `now`, the test can control the time range queried by `indexPieces`.

pyropy: @bajtos I agree, but it could also be done by populating data with some timestamp (i.e. setting 2 deals with `last_payload_retrieval` set to 3 days ago), similar how it's done in some other tests we have.

backend/lib/piece-indexer.js
```diff
@@ -14,25 +16,27 @@ import { getMinerPeerId } from './rpc-service/service.js'
...
    * @param {Queryable} pgPool
    * @param {number} maxDeals
+   * @param {Date} now
    * @returns {Promise<Array<Static< typeof ActiveDealDbEntry>>>}
    */
-export async function fetchDealsWithNoPayloadCid (pgPool, maxDeals) {
-  const query = 'SELECT * FROM active_deals WHERE payload_cid IS NULL ORDER BY activated_at_epoch ASC LIMIT $1'
-  return await loadDeals(pgPool, query, [maxDeals])
+export async function fetchDealsWithNoPayloadCid (pgPool, maxDeals, now) {
+  const query = 'SELECT * FROM active_deals WHERE payload_cid IS NULL AND payload_unretrievable IS DISTINCT FROM TRUE AND (last_payload_retrieval IS NULL OR last_payload_retrieval < $1) ORDER BY activated_at_epoch ASC LIMIT $2'
```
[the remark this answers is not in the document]
NikolasHaimerl: The passing of the now parameter is intentional so it inserted deals can be tested. Using the NOW() operator makes it very hard to test deterministically.

db/migrations/009.do.add-payload-retrievability-column.sql
NikolasHaimerl: I have yes, I think it is more intuitive with the full description. `TIMESTAMPZ` is not a word that is usually used and thus not have any real meaning. Using the full description makes it self explainatory.

backend/lib/piece-indexer.js
NikolasHaimerl: I want to test the inserting of the piece indexer, inserting deals manually would defeat that purpose. I want to assert what timestamp exactly is used where.

db/migrations/009.do.add-payload-retrievability-column.sql
NikolasHaimerl: I used this conciously because the behaviour we want to test is whether a payload is unretrievable. Since null evaluates to False in Javascript an easy mistake could be to ask for all non retrievable payloads and use False somewhere to evaluate what is non retrievable, accidently also fetching all Null values. Using the opposite boolean construction mitigates this pitfall.

backend/lib/piece-indexer.js
```diff
@@ -14,25 +16,27 @@ import { getMinerPeerId } from './rpc-service/service.js'
...
     const minerPeerId = await getMinerPeerId(deal.miner_id, makeRpcRequest)
-    const payloadCid = await getDealPayloadCid(minerPeerId, deal.piece_cid)
-    if (payloadCid) {
-      deal.payload_cid = payloadCid
-      await updatePayloadInActiveDeal(pgPool, deal)
+    deal.payload_cid = await getDealPayloadCid(minerPeerId, deal.piece_cid)
+    if (!deal.payload_cid && deal.last_payload_retrieval) {
+      deal.payload_unretrievable = true
     }
+    deal.last_payload_retrieval = new Date(now)
```
juliangruber, line 26:
```suggestion
```

As discussed in https://github.com/CheckerNetwork/spark-deal-observer/pull/77/files#r1941361889

juliangruber (changes requested):

backend/lib/piece-indexer.js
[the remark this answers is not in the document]
NikolasHaimerl: I wanted to be sure I also coverted the case if `payload_unretrievable` is NULL

backend/lib/piece-indexer.js
NikolasHaimerl: I would disagree, I would much rather have timestamps set that I can controll during testing, rather than have the update function set a timestamp which I then cannot accurately test.

backend/lib/piece-indexer.js
NikolasHaimerl: I created an [issue](https://github.com/CheckerNetwork/spark-deal-observer/issues/86) for this.

backend/lib/piece-indexer.js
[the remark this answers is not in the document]
juliangruber: See https://www.postgresql.org/docs/current/functions-comparison.html:

> NULL::boolean IS NOT TRUE → t (rather than NULL)

I think it also covers it, no?

backend/lib/piece-indexer.js
juliangruber: If you don't control timestamps, you can still test it:

```
const timestampBefore = await pg.query('...')
updatePayloadInActiveDeal(...)
const timestampAfter = await pg.query('...')
assert(timestampAfter > timestampBefore)
```

NikolasHaimerl: What I am only trying to test is not `updatePayloadInActiveDeal` but`lookUpPayloadCids` which is why I am passing the now timestamp to `updatePayloadInActiveDeal`. 
I think it is not very intuitive to pass a deal with certain parameters (including the timestamp of the last lookup) and then have only one field change during insertion time. 

if the clock of NodeJs is off, wouldnt that yield problems throughout all of our repositories? 
We usually do timestamp comparisons in node js, not postgres, so I would go with the time measurement where we do most of our comparisons, no?

backend/lib/deal-observer.js
```diff
@@ -79,7 +79,9 @@ export async function storeActiveDeals (activeDeals, pgPool) {
           term_start_epoch,
           term_min,
           term_max,
-          sector_id
+          sector_id,
+          payload_retrievability_state,
```
juliangruber, line 83:
```suggestion
          payload_retrievability,
```

(everything stored in Postgres is a state)

If you want to annotate the type, maybe instead this?

```suggestion
          payload_retrievability_enum,
```

I suggest we wait for @bajtos with this one

db/migrations/009.do.add-payload-retrievability-column.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TYPE PAYLOAD_RETRIEVABILITY_STATE AS ENUM ('PAYLOAD_CID_NOT_QUERIED_YET', 'PAYLOAD_CID_UNRESOLVED', 'PAYLOAD_CID_RESOLVED','PAYLOAD_CID_TERMINALLY_UNRETRIEVABLE');
```
juliangruber, line 1:
```suggestion
CREATE TYPE PAYLOAD_RETRIEVABILITY_STATE AS ENUM ('PAYLOAD_CID_RESOLUTION_NOT_ATTEMPTED_YET', 'PAYLOAD_CID_UNRESOLVED', 'PAYLOAD_CID_RESOLVED','PAYLOAD_CID_TERMINALLY_UNRETRIEVABLE');
```

juliangruber: So as not to introduce another concept of "querying"

db/lib/types.js
```diff
@@ -1,4 +1,14 @@
 import { Type } from '@sinclair/typebox'
+
+const PayloadRetrievabilityState = {
+  NotQueried: 'PAYLOAD_CID_NOT_QUERIED_YET',
+  Unresolved: 'PAYLOAD_CID_UNRESOLVED',
+  Resolved: 'PAYLOAD_CID_RESOLVED',
+  TerminallyUnretrievable: 'PAYLOAD_CID_TERMINALLY_UNRETRIEVABLE'
+}
+
+const PayloadRetrievabilityStateType = Type.Enum(PayloadRetrievabilityState)
```
juliangruber, line 10:
```suggestion
const PayloadRetrievabilityStates = {
  NotQueried: 'PAYLOAD_CID_NOT_QUERIED_YET',
  Unresolved: 'PAYLOAD_CID_UNRESOLVED',
  Resolved: 'PAYLOAD_CID_RESOLVED',
  TerminallyUnretrievable: 'PAYLOAD_CID_TERMINALLY_UNRETRIEVABLE'
}

const PayloadRetrievabilityState = Type.Enum(PayloadRetrievabilityStates)
```

Does this make more sense? Payload Retrievability States lists all possible states. Payload Retrievability State is one of these states at a time.

We might want to wait for @bajtos with this one

juliangruber (changes requested):

backend/lib/look-up-payload-cids.js
```diff
@@ -14,28 +17,36 @@ import { getMinerPeerId } from './rpc-service/service.js'
...
-    const payloadCid = await getDealPayloadCid(minerPeerId, deal.piece_cid)
-    if (payloadCid) {
-      deal.payload_cid = payloadCid
-      await updatePayloadCidInActiveDeal(pgPool, deal)
+    deal.payload_cid = await getDealPayloadCid(minerPeerId, deal.piece_cid)
+    if (!deal.payload_cid) {
+      if (deal.last_payload_retrieval_attempt) {
+        deal.payload_retrievability_state = PayloadRetrievabilityState.TerminallyUnretrievable
+      } else {
+        deal.payload_retrievability_state = PayloadRetrievabilityState.Unresolved
```
juliangruber, line 29: This is so nice to read :)

backend/lib/look-up-payload-cids.js
```diff
@@ -14,28 +17,36 @@ import { getMinerPeerId } from './rpc-service/service.js'
...
-  for (const deal of await fetchDealsWithUnresolvedPayloadCid(pgPool, maxDeals)) {
+  for (const deal of await fetchDealsWithUnresolvedPayloadCid(pgPool, maxDeals, new Date(now - THREE_DAYS_IN_MILLISECONDS))) {
     const minerPeerId = await getMinerPeerId(deal.miner_id, makeRpcRequest)
-    const payloadCid = await getDealPayloadCid(minerPeerId, deal.piece_cid)
-    if (payloadCid) {
-      deal.payload_cid = payloadCid
-      await updatePayloadCidInActiveDeal(pgPool, deal)
+    deal.payload_cid = await getDealPayloadCid(minerPeerId, deal.piece_cid)
+    if (!deal.payload_cid) {
+      if (deal.last_payload_retrieval_attempt) {
```
juliangruber, line 26:
```suggestion
      if (deal.payload_retrievabillity_state === PayloadRetrievabilityState.Unresolved) {
```

juliangruber (approved):

juliangruber:
@bajtos I'm merging now so that we can hopefully get more payloads assigned over the weekend, and thereby improve Spark DDO data

db/lib/types.js
bajtos: Your suggestion makes sense to me.

backend/lib/look-up-payload-cids.js
```diff
@@ -14,28 +17,36 @@ import { getMinerPeerId } from './rpc-service/service.js'
...
    * @param {Queryable} pgPool
    * @param {number} maxDeals
+   * @param {Date} now
    * @returns {Promise<Array<Static< typeof ActiveDealDbEntry>>>}
    */
-async function fetchDealsWithUnresolvedPayloadCid (pgPool, maxDeals) {
-  const query = 'SELECT * FROM active_deals WHERE payload_cid IS NULL ORDER BY activated_at_epoch ASC LIMIT $1'
-  return await loadDeals(pgPool, query, [maxDeals])
+export async function fetchDealsWithUnresolvedPayloadCid (pgPool, maxDeals, now) {
+  const query = "SELECT * FROM active_deals WHERE payload_cid IS NULL AND (payload_retrievability_state = 'PAYLOAD_CID_NOT_QUERIED_YET' OR payload_retrievability_state = 'PAYLOAD_CID_UNRESOLVED') AND (last_payload_retrieval_attempt IS NULL OR last_payload_retrieval_attempt < $1) ORDER BY activated_at_epoch ASC LIMIT $2"
```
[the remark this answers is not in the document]
juliangruber: I also find it difficult to read 👍 But, this helps prioritizing the linter work even more. I added https://github.com/CheckerNetwork/roadmap/issues/166 as candidate to https://github.com/CheckerNetwork/roadmap/issues/235

db/migrations/009.do.add-payload-retrievability-column.sql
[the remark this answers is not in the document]
juliangruber: great point, I missed here that we made the same mistake again of calling "payload cid" a "payload".

PR #89, opened 2025-02-07
Fix observe actor events telemetry name

juliangruber:


backend/bin/deal-observer-backend.js
```diff
@@ -57,10 +56,10 @@ const observeActorEventsLoop = async (makeRpcRequest, pgPool) => {
       Sentry.captureException(e)
     }
     const dt = Date.now() - start
-    console.log(`Loop "${LOOP_NAME}" took ${dt}ms`)
+    console.log(`Loop "Observe actor events" took ${dt}ms`)
 
     if (INFLUXDB_TOKEN) {
-      recordTelemetry(`loop_${slug(LOOP_NAME, '_')}`, point => {
+      recordTelemetry('loop_builtin_actor_events', point => {
```
juliangruber, line 62: Changing this to "loop_observe_actor_events" was unintentional, @NikolasHaimerl and me agreed that the `LOOP_NAME` is risky

NikolasHaimerl:
@juliangruber  should we also do this change for the other loops that record the loop time?

NikolasHaimerl (approved):

backend/bin/deal-observer-backend.js
```diff
@@ -57,10 +55,10 @@ const observeActorEventsLoop = async (makeRpcRequest, pgPool) => {
       Sentry.captureException(e)
     }
     const dt = Date.now() - start
-    console.log(`Loop "${LOOP_NAME}" took ${dt}ms`)
+    console.log(`Loop "Observe actor events" took ${dt}ms`)
```
bajtos, line 58: I propose the following log for consistency with the telemetry point name.

```suggestion
    console.log(`Loop "Observe built-in actor events" took ${dt}ms`)
```

backend/bin/deal-observer-backend.js
```diff
@@ -103,10 +100,10 @@ const sparkApiSubmitDealsLoop = async (pgPool, { sparkApiBaseUrl, sparkApiToken,
       Sentry.captureException(e)
     }
     const dt = Date.now() - start
-    console.log(`Loop "${LOOP_NAME}" took ${dt}ms`)
+    console.log(`Loop "Submit deals to spark-api" took ${dt}ms`)
 
     if (INFLUXDB_TOKEN) {
-      recordTelemetry(`loop_${slug(LOOP_NAME, '_')}`, point => {
+      recordTelemetry('loop_submit_deals_to_sparkapi', point => {
```
bajtos, line 106:
```suggestion
      recordTelemetry('loop_submit_deals_to_spark_api', point => {
```

(unless you want to preserve the point name)

bajtos (approved):
LGTM with a few suggestions to consider.

FWIW, the slug-based implementation comes from [spark-stats/spark-observer](https://github.com/CheckerNetwork/spark-stats/blob/ff6b3d21848b4aef46cc1a29a8a02bc9a0c96fc3/observer/bin/spark-observer.js#L59-L75), where we have a shared helper method to repeatedly call the same function with a delay between iterations.

https://github.com/CheckerNetwork/spark-stats/blob/ff6b3d21848b4aef46cc1a29a8a02bc9a0c96fc3/observer/bin/spark-observer.js#L59-L75

backend/bin/deal-observer-backend.js
juliangruber: that was the plan. Do you think it's worth it to make a breaking change in the point name?

juliangruber: The whole point of the PR is to keep existing telemetry names in tact, so I'll keep this one as well

PR #90, opened 2025-02-10
fix: consistent naming for resolving payloads

### LINKED ISSUE #87: Consistent naming for resolving payload CIDs
I think what you're trying to say is that we can rename `lookUp..()` to `resolve..()` or something similar in a follow-up PR?

_Originally posted by @juliangruber in https://github.com/CheckerNetwork/spark-deal-observer/pull/84#discussion_r1944525332_

NikolasHaimerl:
We want consistent naming for resolving payload CIDs:
The action is `resolving` the state is `resolved` or `unresolved`. 

Closes #87

### DIFF
--- a/backend/bin/deal-observer-backend.js
+++ b/backend/bin/deal-observer-backend.js
@@ -8,9 +8,9 @@ import '../lib/instrument.js'
 import { createInflux } from '../lib/telemetry.js'
 import { rpcRequest } from '../lib/rpc-service/service.js'
 import { fetchDealWithHighestActivatedEpoch, countStoredActiveDeals, observeBuiltinActorEvents } from '../lib/deal-observer.js'
-import { countStoredActiveDealsWithUnresolvedPayloadCid, lookUpPayloadCids } from '../lib/look-up-payload-cids.js'
+import { countStoredActiveDealsWithUnresolvedPayloadCid, resolvePayloadCids } from '../lib/resolve-payload-cids.js'
 import { findAndSubmitUnsubmittedDeals, submitDealsToSparkApi } from '../lib/spark-api-submit-deals.js'
-import { getDealPayloadCid } from '../lib/piece-indexer-service.js'
+import { resolvePayloadCid } from '../lib/piece-indexer-service.js'
 /** @import {Queryable} from '@filecoin-station/deal-observer-db' */
 
 const {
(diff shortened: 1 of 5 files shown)


backend/bin/deal-observer-backend.js
```diff
@@ -117,19 +117,19 @@ const sparkApiSubmitDealsLoop = async (pgPool, { sparkApiBaseUrl, sparkApiToken,
   }
 }
 
-export const lookUpPayloadCidsLoop = async (makeRpcRequest, getDealPayloadCid, pgPool) => {
-  const LOOP_NAME = 'Look up payload CIDs'
+export const resolvePayloadCidsLoop = async (makeRpcRequest, resolvePayloadCid, pgPool) => {
+  const LOOP_NAME = 'Resolve payload CIDs'
```
NikolasHaimerl, line 121: We will have to change the Grafana query after this change.

backend/bin/deal-observer-backend.js
```diff
@@ -117,19 +117,19 @@ const sparkApiSubmitDealsLoop = async (pgPool, { sparkApiBaseUrl, sparkApiToken,
...
-    // Maximum number of deals to look up payload CIDs for in one loop iteration
+    // Maximum number of deals to resolve payload CIDs for in one loop iteration
     const maxDeals = 1000
     try {
-      const numOfPayloadCidsResolved = await lookUpPayloadCids(makeRpcRequest, getDealPayloadCid, pgPool, maxDeals)
+      const numOfPayloadCidsResolved = await resolvePayloadCids(makeRpcRequest, resolvePayloadCid, pgPool, maxDeals)
       const numOfUnresolvedPayloadCids = await countStoredActiveDealsWithUnresolvedPayloadCid(pgPool)
       if (INFLUXDB_TOKEN) {
-        recordTelemetry('look_up_payload_cids_stats', point => {
+        recordTelemetry('resolve_payload_cids_stats', point => {
```
NikolasHaimerl, line 130: Will need to change the Grafana query after this change

backend/lib/piece-indexer-service.js
```diff
@@ -30,7 +30,7 @@ export const getDealPayloadCid = async (providerId, pieceCid) => {
       if (parsedPixResponse.error === 'PROVIDER_OR_PIECE_NOT_FOUND') {
         return null
       }
-    } catch {}
+    } catch { }
```
pyropy, line 33: Should we throw or log error here?

pyropy (approved):
Great job! 👏🏻

backend/bin/deal-observer-backend.js
```diff
@@ -8,9 +8,9 @@ import '../lib/instrument.js'
 import { createInflux } from '../lib/telemetry.js'
 import { rpcRequest } from '../lib/rpc-service/service.js'
 import { fetchDealWithHighestActivatedEpoch, countStoredActiveDeals, observeBuiltinActorEvents } from '../lib/deal-observer.js'
-import { countStoredActiveDealsWithUnresolvedPayloadCid, lookUpPayloadCids } from '../lib/look-up-payload-cids.js'
+import { countStoredActiveDealsWithUnresolvedPayloadCid, resolvePayloadCids } from '../lib/resolve-payload-cids.js'
 import { findAndSubmitUnsubmittedDeals, submitDealsToSparkApi } from '../lib/spark-api-submit-deals.js'
-import { getDealPayloadCid } from '../lib/piece-indexer-service.js'
+import { resolvePayloadCid } from '../lib/piece-indexer-service.js'
```
juliangruber, line 13: This is confusing: One function is called `resolvePayloadCids`, the other `resolvePayloadCid`. Looking at the name, I would expect the 1st to do the same as the 2nd, only for multiple payload CIDs.

It's different from that though: The 1st looks up what to resolve, calls the look up function, and also saves the state in the DB.

I don't have a solution on the top of my head. @bajtos wdyt?

backend/lib/piece-indexer-service.js
juliangruber: No, as this is checking if the response is an error. If it's not, we parse the response again, below (where we would throw)

juliangruber (changes requested):

backend/bin/deal-observer-backend.js
juliangruber: @NikolasHaimerl do you have a suggestion? Let's use "resolve" for one of the two cases from the previous comment, and another term for the other

NikolasHaimerl: You are suggesting we use two different verbs, one for the action of fetching the payloads via HTTP and another for the overall process of filling payload CIDs in the database that are currently set to NULL. Did I understand that correctly?

juliangruber: What I'm saying is that this is currently inconsistent naming. Consistent naming means to give the same thing the same name, but in turn, to give different things different names. Here we have two processes (currently called `resolvePayloadCid` and `resolvePayloadCids`), that are different (in more ways than their multitude). Therefore they need different names.

NikolasHaimerl: I'll use the same naming we are already using for rpc requests --> `payloadCidRequest`/`rpcRequest` and `makePayloadCidRequest`/`makeRpcRequest`

juliangruber: Great thinking!

juliangruber (approved):
Thanks for finding consistent naming 🙏

PR #92, opened 2025-02-10
chore: add serial id

### LINKED ISSUE #86: Column Uniqueness for active deals
## Description
Currently, updating individual active deal rows requires complex WHERE conditions that compare values across multiple columns (9 columns as mentioned in the discussion). This makes queries verbose and potentially less efficient.

## Proposed Solution
Add an auto-generated primary key to the relevant table(s) so that we can simplify WHERE conditions from:
```sql
WHERE activated_at_epoch = $4 AND miner_id = $5 AND client_id = $6 AND piece_cid = $7 AND ...
```
to:
```sql
WHERE id = $4
```

## Benefits
- Simplified queries
- Potentially improved performance for updates
- Clearer code maintenance
- More direct way to reference unique rows

## Implementation Considerations
- Need to ensure uniqueness in the database is properly maintained
- Will require updating existing queries that rely on the multi-column WHERE clauses

_Originally posted by @bajtos in https://github.com/CheckerNetwork/spark-deal-observer/pull/77#discussion_r1942489012_

NikolasHaimerl:
The table `active_deals` does not have a unique property right now. This makes it tedious to fetch deals as one needs to check each field to a specific value to retrieve it. We can make this shorter by introducing a serial and set it as the primary key of the table `active_deals`. Consequently, we only need to ask for the primary key if we want to fetch or update a certain deal. 

This PR adds a serial to the active deals table which is then set as the primary key with the column name `id`. 
Closes #86


pyropy (changes requested):
We would also need to update parts in spark deal submitter where we look for [unsubmitted deals](https://github.com/CheckerNetwork/spark-deal-observer/blob/bda864b8c738789c121da9b21be127ebe8e2a3f2/backend/lib/spark-api-submit-deals.js#L56-L62) and [mark deals as submitted](https://github.com/CheckerNetwork/spark-deal-observer/blob/bda864b8c738789c121da9b21be127ebe8e2a3f2/backend/lib/spark-api-submit-deals.js#L81-L108) before merging this.

NikolasHaimerl:
> We would also need to update parts in spark deal submitter where we look for [unsubmitted deals](https://github.com/CheckerNetwork/spark-deal-observer/blob/bda864b8c738789c121da9b21be127ebe8e2a3f2/backend/lib/spark-api-submit-deals.js#L56-L62) and [mark deals as submitted](https://github.com/CheckerNetwork/spark-deal-observer/blob/bda864b8c738789c121da9b21be127ebe8e2a3f2/backend/lib/spark-api-submit-deals.js#L81-L108) before merging this.

Is there a certain reason we did not use `ActiveDealDbEntry` for the type of `eligibleDeals` and the return type of the generator for `findUnsubmittedDeals`?

pyropy:
> > We would also need to update parts in spark deal submitter where we look for [unsubmitted deals](https://github.com/CheckerNetwork/spark-deal-observer/blob/bda864b8c738789c121da9b21be127ebe8e2a3f2/backend/lib/spark-api-submit-deals.js#L56-L62) and [mark deals as submitted](https://github.com/CheckerNetwork/spark-deal-observer/blob/bda864b8c738789c121da9b21be127ebe8e2a3f2/backend/lib/spark-api-submit-deals.js#L81-L108) before merging this.
> 
> Is there a certain reason we did not use `ActiveDealDbEntry` for the type of `eligibleDeals` and the return type of the generator for `findUnsubmittedDeals`?

No there is not, I guess that should be also added.

db/lib/types.js
```diff
@@ -10,6 +10,7 @@ const PayloadRetrievabilityState = {
 const PayloadRetrievabilityStateType = Type.Enum(PayloadRetrievabilityState)
 
 const ActiveDealDbEntry = Type.Object({
+  id: Type.Optional(Type.Number()),
```
juliangruber, line 13: What would be missing to make this type required (which should be safer)?

juliangruber: Am I right in my understanding that we need this to be optional so that when we create a new entry from the block event (in backend/lib/utils.js) we can keep it empty?

backend/lib/look-up-payload-cids.js
```diff
@@ -67,22 +67,14 @@ async function updatePayloadCidInActiveDeal (pgPool, deal, newPayloadRetrievalSt
...
-      deal.activated_at_epoch,
-      deal.miner_id,
-      deal.client_id,
-      deal.piece_cid,
-      deal.piece_size,
-      deal.term_start_epoch,
-      deal.term_min,
-      deal.term_max,
-      deal.sector_id
+      deal.id
```
juliangruber, line 77: Here we're updating the UPDATE query, however I'm not seeing a matching SELECT query update. Are we using a SELECT *, or is that part missing?

juliangruber (changes requested):

juliangruber:
> We would also need to update parts in spark deal submitter where we look for [unsubmitted deals](https://github.com/CheckerNetwork/spark-deal-observer/blob/bda864b8c738789c121da9b21be127ebe8e2a3f2/backend/lib/spark-api-submit-deals.js#L56-L62) and [mark deals as submitted](https://github.com/CheckerNetwork/spark-deal-observer/blob/bda864b8c738789c121da9b21be127ebe8e2a3f2/backend/lib/spark-api-submit-deals.js#L81-L108) before merging this.

I believe this is what I discovered in https://github.com/CheckerNetwork/spark-deal-observer/pull/92#discussion_r1949154497 as well. Can you please add a test @NikolasHaimerl that shows that this fails, and then follow up with a fix?

db/lib/types.js
bajtos: If we want to be strict about types, then there should be two types:
1. "NewActiveDealDbEntry" that does not have any ID value - we want the database to generate the value for us
2. "ActiveDealDbEntry" where the ID value is required (always present)

bajtos: In TypeScript, it's possible to create a new type from another type by omitting certain properties - see https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys.

```ts
type NewActiveDealDbEntry = Omit<ActiveDealDbEntry, "id">
```

That obviously does not provide runtime validation, so if we want to go down the route of two types, then we need to find a solution based on Typebox features.

NikolasHaimerl: @juliangruber you are correct, if we want to have the type clarity before inserting deals into the database (like we are doing in `backend/lib/utils.js`), we need to keep the id field optional. We do not know the id of a deal before inserting it. The id also does not have any meaning outside of the context of the database. 
The only scenario I see where this could become an issue is if we try to look for a deal using the id and the typing from `Typebox` without knowing the id. 
I do not think it makes much of a difference whether we use typebox during runtime to assert the existence of the id or whether we simply check it using an if statement wherever necessary.

NikolasHaimerl:
> We would also need to update parts in spark deal submitter where we look for [unsubmitted deals](https://github.com/CheckerNetwork/spark-deal-observer/blob/bda864b8c738789c121da9b21be127ebe8e2a3f2/backend/lib/spark-api-submit-deals.js#L56-L62) and [mark deals as submitted](https://github.com/CheckerNetwork/spark-deal-observer/blob/bda864b8c738789c121da9b21be127ebe8e2a3f2/backend/lib/spark-api-submit-deals.js#L81-L108) before merging this.

I am not sure I understand why we would need to make a change here? 
The way I understand it is that we are omitting the `id` along with a view other fields when querying for submittable deals. What change would you make in that regard?

backend/lib/look-up-payload-cids.js
juliangruber: https://github.com/CheckerNetwork/spark-deal-observer/blob/bda864b8c738789c121da9b21be127ebe8e2a3f2/backend/lib/look-up-payload-cids.js#L48

We are using `SELECT *`, so indeed all deals fetched from the database will have an `.id` property now ✅

db/lib/types.js
juliangruber: > if we want to have the type clarity before inserting deals into the database (like we are doing in backend/lib/utils.js), we need to keep the id field optional. 

On the contrary, the optional id field does not give type clarity. If we wanted to have type clarify, we would need separate types, one for with ID and one for without, so that at any time you know with certainty whether an object has the ID or not.

I'm fine not adding a second type, if we find a better name for the type. The type is currently called `ActiveDealDbEntry`. Such a type must have an `id`, as we can't have a db entry without an id (due to the schema). Therefore, this type gives a false sense of security.

NikolasHaimerl: I see your point. Let me split it up then into a deal type before inserting into the database and a deal type that corresponds to the data being read from the database.

bajtos: I see you added `ActiveDeal` type for data without the auto-generated `id` property 👍🏻 

I would expect the `id` property of `ActiveDealDbEntry` to be required now. Do you agree?

```suggestion
  id: Type.Number(),
```

backend/test/look-up-payload-cids.test.js
```diff
@@ -8,7 +8,7 @@ import assert from 'assert'
 import { minerPeerIds } from './test_data/minerInfo.js'
 import { payloadCIDs } from './test_data/payloadCIDs.js'
 import { Value } from '@sinclair/typebox/value'
-import { ActiveDealDbEntry, PayloadRetrievabilityState } from '@filecoin-station/deal-observer-db/lib/types.js'
+import { ActiveDealDbEntry as ActiveDeal, PayloadRetrievabilityState } from '@filecoin-station/deal-observer-db/lib/types.js'
```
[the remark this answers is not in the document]
NikolasHaimerl: Yes, thanks for catching that. It was created by the IDE indeed.

backend/test/look-up-payload-cids.test.js
```diff
@@ -128,14 +130,15 @@ describe('deal-observer-backend piece indexer payload retrieval', () => {
     })
 
     await storeActiveDeals([deal], pgPool)
-    assert.deepStrictEqual((await loadDeals(pgPool, 'SELECT * FROM active_deals'))[0], deal)
+    const dealDbEntry = { id: 1, ...deal }
+    assert.deepStrictEqual((await loadDeals(pgPool, 'SELECT * FROM active_deals'))[0], dealDbEntry)
```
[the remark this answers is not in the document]
NikolasHaimerl: The intention of this assertion is to verify that the expected deal is stored in the database.

backend/test/look-up-payload-cids.test.js
```diff
@@ -128,14 +130,15 @@ describe('deal-observer-backend piece indexer payload retrieval', () => {
...
     // The payload is unretrievable and the last retrieval timestamp should be updated
     await lookUpPayloadCids(fetchMinerId, getDealPayloadCid, pgPool, 10000, now)
     // The timestamp on when the last retrieval of the payload was, was not yet set, so the piece indexer will try to fetch the payload
     assert.strictEqual(payloadsCalled, 1)
-    deal.last_payload_retrieval_attempt = new Date(now)
-    deal.payload_retrievability_state = PayloadRetrievabilityState.Unresolved
-    assert.deepStrictEqual((await loadDeals(pgPool, 'SELECT * FROM active_deals'))[0], deal)
+    dealDbEntry.last_payload_retrieval_attempt = new Date(now)
+    dealDbEntry.payload_retrievability_state = PayloadRetrievabilityState.Unresolved
+    assert.deepStrictEqual((await loadDeals(pgPool, 'SELECT * FROM active_deals'))[0], dealDbEntry)
```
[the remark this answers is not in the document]
juliangruber: What we have now is like snapshot testing, which is useful as well - it rules out that a patch is accidentally changing a property that you didn't mean to. Therefore it's safer than just testing individual updates.

The downsides as you mentioned are that it's lacking specificity, and failures are harder to interpret. Another downside is that if you add a property to the schema, all tests need to be updated.

I don't see an easy way to decide what is the right approach here.

pyropy:
> I am not sure I understand why we would need to make a change here? The way I understand it is that we are omitting the `id` along with a few other fields when querying for submittable deals. What change would you make in that regard?

After we submit the deal to spark-api we [update active deals](https://github.com/CheckerNetwork/spark-deal-observer/blob/b845b280d4925b2b6b6d08934701da3362ab6615/backend/lib/spark-api-submit-deals.js#L81-L108) with `submitted_at` timestamp (marking them as submitted). Updating the records is done the same way we update records with `payload_cid` and introduction of `id` would/will solve that issue.

backend/test/look-up-payload-cids.test.js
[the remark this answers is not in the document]
NikolasHaimerl: @bajtos is this something we could outsource into its own ticket?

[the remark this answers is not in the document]
bajtos: After reading @juliangruber's comment, I can see how the current tests can provide different benefits than what I described in my comment.

Let's keep the test codes you have right now and revisit this topic if it ever becomes a problem in the future.

backend/test/look-up-payload-cids.test.js
```diff
@@ -168,9 +171,10 @@ describe('deal-observer-backend piece indexer payload retrieval', () => {
     await lookUpPayloadCids(fetchMinerId, getDealPayloadCid, pgPool, 10000, now)
     assert.strictEqual(payloadsCalled, 1)
     // This is the second attempt that failed to fetch the payload CID so the deal should be marked as unretrievable
-    deal.payload_retrievability_state = PayloadRetrievabilityState.TerminallyUnretrievable
-    deal.last_payload_retrieval_attempt = new Date(now)
-    assert.deepStrictEqual((await loadDeals(pgPool, 'SELECT * FROM active_deals'))[0], deal)
+    const expectedDealDbEntry = { id: 1, ...deal }
+    expectedDealDbEntry.payload_retrievability_state = PayloadRetrievabilityState.TerminallyUnretrievable
+    expectedDealDbEntry.last_payload_retrieval_attempt = new Date(now)
```
bajtos, line 176: Nitpick, feel free to ignore.

This can be written as follows:

```js
const expectedDealDbEntry = { 
  id: 1, 
  ...deal,
  payload_retrievability_state: PayloadRetrievabilityState.TerminallyUnretrievable,
  last_payload_retrieval_attempt: new Date(now)
}
```

backend/test/look-up-payload-cids.test.js
```diff
@@ -202,11 +206,12 @@ describe('deal-observer-backend piece indexer payload retrieval', () => {
     await storeActiveDeals([deal], pgPool)
     await lookUpPayloadCids(fetchMinerId, getDealPayloadCid, pgPool, 10000, now)
     assert.strictEqual(payloadsCalled, 1)
-    deal.last_payload_retrieval_attempt = new Date(now)
-    deal.payload_cid = payloadCid
-    deal.payload_retrievability_state = PayloadRetrievabilityState.Resolved
+    const expectedDealDbEntry = { id: 1, ...deal }
+    expectedDealDbEntry.last_payload_retrieval_attempt = new Date(now)
+    expectedDealDbEntry.payload_cid = payloadCid
+    expectedDealDbEntry.payload_retrievability_state = PayloadRetrievabilityState.Resolved
```
bajtos, line 212: Ditto.

bajtos (commented):
LGTM 👏🏻 

Please consider addressing https://github.com/CheckerNetwork/spark-deal-observer/pull/92#issuecomment-2654207606 in a follow-up pull request after landing this one, to finish the simplification work enabled by the `id` PK.

bajtos (approved):

NikolasHaimerl:
> > I am not sure I understand why we would need to make a change here? The way I understand it is that we are omitting the `id` along with a few other fields when querying for submittable deals. What change would you make in that regard?
> 
> After we submit the deal to spark-api we [update active deals](https://github.com/CheckerNetwork/spark-deal-observer/blob/b845b280d4925b2b6b6d08934701da3362ab6615/backend/lib/spark-api-submit-deals.js#L81-L108) with `submitted_at` timestamp (marking them as submitted). Updating the records is done the same way we update records with `payload_cid` and introduction of `id` would/will solve that issue.

Eligible deals does not have an id in it, no? See https://github.com/CheckerNetwork/spark-deal-observer/blob/main/backend/lib/spark-api-submit-deals.js#L154

pyropy:
> > > I am not sure I understand why we would need to make a change here? The way I understand it is that we are omitting the `id` along with a few other fields when querying for submittable deals. What change would you make in that regard?
> > 
> > 
> > After we submit the deal to spark-api we [update active deals](https://github.com/CheckerNetwork/spark-deal-observer/blob/b845b280d4925b2b6b6d08934701da3362ab6615/backend/lib/spark-api-submit-deals.js#L81-L108) with `submitted_at` timestamp (marking them as submitted). Updating the records is done the same way we update records with `payload_cid` and introduction of `id` would/will solve that issue.
> 
> Eligible deals does not have an id in it, no? See https://github.com/CheckerNetwork/spark-deal-observer/blob/main/backend/lib/spark-api-submit-deals.js#L154

We can always map `ActiveDeal` into `SubmittableDeal` before the submission, not during the select. `SubmittableDeal` is just a type which makes sure that we have correct fields for submission to `spark-api` and is not persisted in some other table (as part of the deal-observer) as such.

That would also means that `expires_at` would be computed in afterwards, and not in the database.

NikolasHaimerl:
> > > > I am not sure I understand why we would need to make a change here? The way I understand it is that we are omitting the `id` along with a few other fields when querying for submittable deals. What change would you make in that regard?
> > > 
> > > 
> > > After we submit the deal to spark-api we [update active deals](https://github.com/CheckerNetwork/spark-deal-observer/blob/b845b280d4925b2b6b6d08934701da3362ab6615/backend/lib/spark-api-submit-deals.js#L81-L108) with `submitted_at` timestamp (marking them as submitted). Updating the records is done the same way we update records with `payload_cid` and introduction of `id` would/will solve that issue.
> > 
> > 
> > Eligible deals does not have an id in it, no? See https://github.com/CheckerNetwork/spark-deal-observer/blob/main/backend/lib/spark-api-submit-deals.js#L154
> 
> We can always map `ActiveDeal` into `SubmittableDeal` before the submission, not during the select. `SubmittableDeal` is just a type which makes sure that we have correct fields for submission to `spark-api` and is not persisted in some other table (as part of the deal-observer) as such.
> 
> That would also means that `expires_at` would be computed in afterwards, and not in the database.

I still don't understand why we need to deal with the `id` during the submitting of a deal. We call the function `markDealsAsSubmitted` with a deal returned by `findUnsubmittedDeals`, which both use `SubmittableDeal` and thus have no `id`. Could you elaborate, please?

pyropy:
> > > > > I am not sure I understand why we would need to make a change here? The way I understand it is that we are omitting the `id` along with a few other fields when querying for submittable deals. What change would you make in that regard?
> > > > 
> > > > 
> > > > After we submit the deal to spark-api we [update active deals](https://github.com/CheckerNetwork/spark-deal-observer/blob/b845b280d4925b2b6b6d08934701da3362ab6615/backend/lib/spark-api-submit-deals.js#L81-L108) with `submitted_at` timestamp (marking them as submitted). Updating the records is done the same way we update records with `payload_cid` and introduction of `id` would/will solve that issue.
> > > 
> > > 
> > > Eligible deals does not have an id in it, no? See https://github.com/CheckerNetwork/spark-deal-observer/blob/main/backend/lib/spark-api-submit-deals.js#L154
> > 
> > 
> > We can always map `ActiveDeal` into `SubmittableDeal` before the submission, not during the select. `SubmittableDeal` is just a type which makes sure that we have correct fields for submission to `spark-api` and is not persisted in some other table (as part of the deal-observer) as such.
> > That would also means that `expires_at` would be computed in afterwards, and not in the database.
> 
> I still don't understand why we need to deal with the `id` during the submitting of a deal. We call the function `markDealsAsSubmitted` with a deal returned by `findUnsubmittedDeals`, which both use `SubmittableDeal` and thus have no `id`. Could you elaborate, please?

I know it can be confusing since the `SubmittableDeal` lacks `id` fields, and I agree it's not needed inside `SubmittableDeal` type,  but my reasoning is that since we introduced `id` field to `active_deals` table it would be good practice to update `active_deals` returned by `findUnsubmittedDeals` using `id` for lookup instead of doing lookup by fields present inside `SubmittableDeal`. 

I propose following flow:
- `findUnsubmittedDeals` returns unsubmitted `ActiveDeal` array
- `submitDeals` accepts `Array<ActiveDeal>` maps each `ActiveDeal` to `SubmittableDeal` and submits them to `spark-api`
- `markDealsAsSubmitted` accepts `Array<ActiveDeal>` and updates deals based on their `id`

Ultimately it's up to you to make an decision, but I don't see a reason why we wouldn't use `id` field to perform `markDealsAsSubmitted` update since it's more accurate and faster.

pyropy (approved):
Good job 👏🏻 We can address #116 in a follow-up PR.

bajtos:
> > > > > I am not sure I understand why we would need to make a change here? The way I understand it is that we are omitting the `id` along with a few other fields when querying for submittable deals. What change would you make in that regard?
> > > > 
> > > > 
> > > > After we submit the deal to spark-api we [update active deals](https://github.com/CheckerNetwork/spark-deal-observer/blob/b845b280d4925b2b6b6d08934701da3362ab6615/backend/lib/spark-api-submit-deals.js#L81-L108) with `submitted_at` timestamp (marking them as submitted). Updating the records is done the same way we update records with `payload_cid` and introduction of `id` would/will solve that issue.
> > > 
> > > 
> > > Eligible deals does not have an id in it, no? See https://github.com/CheckerNetwork/spark-deal-observer/blob/main/backend/lib/spark-api-submit-deals.js#L154
> > 
> > 
> > We can always map `ActiveDeal` into `SubmittableDeal` before the submission, not during the select. `SubmittableDeal` is just a type which makes sure that we have correct fields for submission to `spark-api` and is not persisted in some other table (as part of the deal-observer) as such.
> > That would also means that `expires_at` would be computed in afterwards, and not in the database.
> 
> I still don't understand why we need to deal with the `id` during the submitting of a deal. 

It's the same motivation you have in this pull request - we want to avoid a complex WHERE condition when marking deals as submitted.

Quoting from the pull request description:

> This makes it tedious to fetch deals as one needs to check each field to a specific value to retrieve it. We can make this shorter by introducing a serial and set it as the primary key of the table `active_deals`. Consequently, we only need to ask for the primary key if we want to fetch or update a certain deal. 

I am fine with leaving that out of the scope of this PR. Just be mindful that this approach is adding work to our future milestones, which means we will have less room in those milestones for other work.

NikolasHaimerl:
@CodiumAI-Agent /review

NikolasHaimerl:
@juliangruber  it seems that I still need your approval because there are changes requested, but I do not see any for this PR. Could I ask you to PTAL?

pyropy:
> @juliangruber it seems that I still need your approval because there are changes requested, but I do not see any for this PR. Could I ask you to PTAL?

We could dismiss Julians review, WDYT? cc @bajtos

NikolasHaimerl:
> > @juliangruber it seems that I still need your approval because there are changes requested, but I do not see any for this PR. Could I ask you to PTAL?
> 
> We could dismiss Julians review, WDYT? cc @bajtos

Ping @bajtos

PR #80, opened 2025-02-04
chore: enable strict type checking

NikolasHaimerl:
This PR proposes the following changes:

1. Enable strict type checking on the deal observer repository

See [issue](https://github.com/CheckerNetwork/spark-deal-observer/issues/51).

PR #95, opened 2025-02-11
add SubmittableDeal type

NikolasHaimerl:
We are currently not using a specific type when submitting deals through the deal observer backend. 
In two recent discussions ([Here](https://github.com/CheckerNetwork/spark-deal-observer/pull/92#issuecomment-2647927720) and [here](https://github.com/CheckerNetwork/spark-deal-observer/pull/80#discussion_r1940940641)), the need for a specific type for submittable deals has come up. 

This PR suggests a new type for deals that are found to be submittable and used throughout the submitting process of the deal observer. 
This solves the issue of the use of the any type when activating type checking and also delivers clarity on the type returned by various functions during the submitting process.

### DIFF
--- a/backend/lib/spark-api-submit-deals.js
+++ b/backend/lib/spark-api-submit-deals.js
@@ -1,13 +1,17 @@
 /** @import {PgPool, Queryable} from '@filecoin-station/deal-observer-db' */
+/** @import { Static } from '@sinclair/typebox' */
+/** @import { ActiveDealDbEntry, PayloadRetrievabilityStateType } from '@filecoin-station/deal-observer-db/lib/types.js' */
 import Cursor from 'pg-cursor'
 import * as Sentry from '@sentry/node'
+import { Type } from '@sinclair/typebox'
+import { Value } from '@sinclair/typebox/value'
 
 /**
  * Finds deals that haven't been submitted to the Spark API yet and submits them.
  *
  * @param {PgPool} pgPool
  * @param {number} batchSize
- * @param {(eligibleDeals: Array) => Promise<{ingested: number; skipped: number}>} submitDeals
+ * @param {(eligibleDeals: Array<Static<typeof SubmittableDeal>>) => Promise<{ingested: number; skipped: number}>} submitDeals
  * @returns {Promise<{submitted: number; ingested: number; skipped: number;}>} Number of deals submitted, ingested and skipped
  */
 export const findAndSubmitUnsubmittedDeals = async (pgPool, batchSize, submitDeals) => {
@@ -45,7 +49,7 @@ export const findAndSubmitUnsubmittedDeals = async (pgPool, batchSize, submitDea
  *
  * @param {PgPool} pgPool
  * @param {number} batchSize
- * @returns {AsyncGenerator<Array>}
+ * @returns {AsyncGenerator<Array<Static<typeof SubmittableDeal>>, void, unknown>}
  */
 const findUnsubmittedDeals = async function * (pgPool, batchSize) {
   const client = await pgPool.connect()
@@ -72,7 +76,7 @@ const findUnsubmittedDeals = async function * (pgPool, batchSize) {
   while (true) {
     const rows = await cursor.read(batchSize)
     if (rows.length === 0) break
-    yield rows
+    yield rows.map(row => Value.Parse(SubmittableDeal, row))
   }
 
   client.release()
@@ -82,7 +86,7 @@ const findUnsubmittedDeals = async function * (pgPool, batchSize) {
  * Mark deals as submitted.
  *
  * @param {Queryable} pgPool
- * @param {Array} eligibleDeals
+ * @param {Array<Static<typeof SubmittableDeal>>} eligibleDeals
  */
 const markDealsAsSubmitted = async (pgPool, eligibleDeals) => {
   await pgPool.query(`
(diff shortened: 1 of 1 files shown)


backend/lib/spark-api-submit-deals.js
```diff
@@ -1,13 +1,17 @@
 /** @import {PgPool, Queryable} from '@filecoin-station/deal-observer-db' */
+/** @import { Static } from '@sinclair/typebox' */
+/** @import { ActiveDealDbEntry, PayloadRetrievabilityStateType } from '@filecoin-station/deal-observer-db/lib/types.js' */
```
bajtos, line 3: These two imports seem unused; can we remove this line?

```suggestion
```

bajtos (approved):
The proposal in this pull request is consistent with what we are doing elsewhere in this project, I don't see any obvious issues besides one minor comment below.

backend/lib/spark-api-submit-deals.js
```diff
@@ -45,7 +49,7 @@ export const findAndSubmitUnsubmittedDeals = async (pgPool, batchSize, submitDea
  *
  * @param {PgPool} pgPool
  * @param {number} batchSize
- * @returns {AsyncGenerator<Array>}
+ * @returns {AsyncGenerator<Array<Static<typeof SubmittableDeal>>, void, unknown>}
```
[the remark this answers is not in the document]
bajtos: Great question, I missed this detail myself.

FWIW, here is how TypeScript defines the AsyncGenerator type:

```ts
interface AsyncGenerator<T = unknown, TReturn = any, TNext = any> extends AsyncIteratorObject<T, TReturn, TNext> {
    // NOTE: 'next' is defined using a tuple to ensure we report the correct assignability errors in all places.
    next(...[value]: [] | [TNext]): Promise<IteratorResult<T, TReturn>>;
    return(value: TReturn | PromiseLike<TReturn>): Promise<IteratorResult<T, TReturn>>;
    throw(e: any): Promise<IteratorResult<T, TReturn>>;
    [Symbol.asyncIterator](): AsyncGenerator<T, TReturn, TNext>;
}
```

backend/lib/spark-api-submit-deals.js
```diff
@@ -82,7 +86,7 @@ const findUnsubmittedDeals = async function * (pgPool, batchSize) {
  * Mark deals as submitted.
  *
  * @param {Queryable} pgPool
- * @param {Array} eligibleDeals
+ * @param {Array<Static<typeof SubmittableDeal>>} eligibleDeals
```
pyropy, line 89:
```suggestion
 * @param {SubmittableDeal[]} eligibleDeals
```

I think we could simplify this type signature and make it more correct. I have tried to break down the previous type signature using Le Chat:

> 1.  typeof SubmittableDeal: This refers to the type of the SubmittableDeal class or object itself. If SubmittableDeal is a class, typeof SubmittableDeal would be the type of the class constructor.
>
> 2. Static<typeof SubmittableDeal>: The Static utility type is not a built-in TypeScript type, but it is often used in libraries to refer to the static side of a class. This means it refers to the properties and methods that are available on the class itself, rather than on instances of the class.
>
> 3. Array<...>: This indicates that the type is an array. The type inside the angle brackets (<...>) specifies the type of the elements in the array.

I think we can safely remove `Static<...>` as we're not referring to a static side of this class. Did you have for goal to make this array immutable or read only? If that's the case it would be more adequate to use `ReadonlyArray` type.

As for difference between `Array<SubmittableDeal>` and `Array<typeof SubmittableDeal>` Le Chat says the following:

> `SubmittableDeal[]` or `Array<SubmittableDeal>` refers to an array of instances of the `SubmittableDeal `class.
> `Array<typeof SubmittableDeal>` refers to an array of the `SubmittableDeal` class constructors or static members.

Given this breakdown I think it would be more correct if we change type to `SubmittableDeal[]` or `Array<SubmittableDeal>`.

backend/lib/spark-api-submit-deals.js
```diff
@@ -112,7 +116,7 @@ const markDealsAsSubmitted = async (pgPool, eligibleDeals) => {
  *
  * @param {string} sparkApiBaseURL
  * @param {string} sparkApiToken
- * @param {Array} deals
+ * @param {Array<Static<typeof SubmittableDeal>>} deals
```
pyropy, line 119:
```suggestion
 * @param {SubmittableDeal[]} deals
```

Same [comment ](https://github.com/CheckerNetwork/spark-deal-observer/pull/95/files#r1952235257) is applicable here.

backend/lib/spark-api-submit-deals.js
pyropy, line 52:
```suggestion
 * @returns {AsyncGenerator<SubmittableDeal[], void, unknown>}
```

I think same [comment ](https://github.com/CheckerNetwork/spark-deal-observer/pull/95/files#r1952235257) is applicable here.

pyropy (changes requested):
I think we can improve type signature to make it simple and more correct.

backend/lib/spark-api-submit-deals.js
NikolasHaimerl: Yes, I would have expected for the linter to pick thisup.

backend/lib/spark-api-submit-deals.js
NikolasHaimerl: We want the use of the fields for a given class to be checked at compile time, which is why we are using the `Static` type. This was suggested by @bajtos in a [PR](https://github.com/CheckerNetwork/spark-deal-observer/pull/40#pullrequestreview-2578631740) a while back. I think it also applies here.

pyropy (approved):
Good job! 👏🏻

backend/lib/spark-api-submit-deals.js
```diff
@@ -147,3 +150,12 @@ export const submitDealsToSparkApi = async (sparkApiBaseURL, sparkApiToken, deal
 
   return /** @type {{ingested: number; skipped: number}} */ (await response.json())
 }
+
+const SubmittableDeal = Type.Object({
```
pyropy, line 154: Would it make sense if this type would be an union type of `ActiveDeal` and type with `expires_at` attribute?

PR #109, opened 2025-02-14
test: handling of duplicate claim events

NikolasHaimerl:
This PR tests the recently observed bug while trying to insert a duplicate deal. 
The hot fix was merged in this [PR](https://github.com/CheckerNetwork/spark-deal-observer/pull/105/files). 
This PR tests this change by inserting two events with a different event id. This event id is not part of the unique constraint of the `active_deals` table and events that only differ in their event id lead to a unique constraint violation. 
The database should handle this by ignoring the duplicate event.


bajtos:
> fix: test unique bug 

The convention is to use `fix:` prefix to bug fixes. Since this pull request is adding a new test, I think `test:` prefix is better. You can learn more about these convention here: https://www.conventionalcommits.org/en/v1.0.0/

We are not enforcing conventional commits in our repos, so you also have the option of not using any prefix.

When I see the name "unique bug", it sounds like the fixed bug is unique. 

Here is a title I would use myself:

> test: handling of duplicate claim events

juliangruber (changes requested):

backend/test/deal-observer.test.js
```diff
@@ -107,6 +107,36 @@ describe('deal-observer-backend', () => {
     await storeBlockEvent(data)
     assert.strictEqual(await countStoredActiveDeals(pgPool), 2n)
   })
+
+  it('deal with duplicate events', async () => {
```
bajtos, line 111: Based on the [discussion on Slack](https://space-meridian.slack.com/archives/CP50PPW2X/p1739464959149629), the events were not duplicate. What happened: a client stored two copies of the same piece in one sector. The events were for different datacap claims (the claim ID was different).

I feel it's important to capture this distinction in the test name.

How about this:

```suggestion
  it('handles one client piece stored twice in the same sector', async () => {
```

backend/test/deal-observer.test.js
```diff
@@ -107,6 +107,36 @@ describe('deal-observer-backend', () => {
...
+    const storeDeal = async (eventData) => {
+      const event = Value.Parse(BlockEvent, { height: 1, event: eventData, emitter: 'f06' })
+      const dbEntry = convertBlockEventToActiveDealDbEntry(event)
+      await storeActiveDeals([dbEntry], pgPool)
+    }
+    const eventData = {
+      id: 1,
```
bajtos, line 118: When we were looking into the problem yesterday, the property name `id` confused me into thinking it's event ID.

It would be great to rename the property to something like `claimId` to make it more clear what kind of ID it is.

Such change is out of the scope of this pull request, though.

backend/test/deal-observer.test.js
```diff
@@ -107,6 +107,36 @@ describe('deal-observer-backend', () => {
...
+      termMax: 12340,
+      sector: 6n,
+      payload_cid: undefined,
+      payload_retrievability_state: PayloadRetrievabilityState.NotQueried,
+      last_payload_retrieval_attempt: undefined
+    }
+    await storeDeal(eventData)
+    let expected = await loadDeals(pgPool, 'SELECT * FROM active_deals')
+    assert.strictEqual(expected.length, 1)
+    // If we only change the id, the unique constraint which does not include the id should prevent the insertion
```
bajtos, line 134:
```suggestion
    // If we only change the id, the unique constraint which does not include the id should not prevent the insertion
```

IIUC, in this test, we verify that the the unique constraint is not violated and the SQL query does not throw an error. As I understand the original comment, it describes the situation before we fixed it yesterday.

bajtos (changes requested):
Thank you for adding the test!

The code looks good at the high level; let's improve details now.

backend/test/deal-observer.test.js
```diff
@@ -107,6 +107,36 @@ describe('deal-observer-backend', () => {
...
+      termStart: 5,
+      termMin: 12340,
+      termMax: 12340,
+      sector: 6n,
+      payload_cid: undefined,
+      payload_retrievability_state: PayloadRetrievabilityState.NotQueried,
+      last_payload_retrieval_attempt: undefined
+    }
+    await storeDeal(eventData)
+    let expected = await loadDeals(pgPool, 'SELECT * FROM active_deals')
```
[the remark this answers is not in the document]
bajtos: Good point. I often use `actual` to store the data we observed. (Vs `expected` to store what we expected to observe.)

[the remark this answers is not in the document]
juliangruber: `actual` is better than `is` 👍

backend/test/deal-observer.test.js
juliangruber: Agreed, this deal stuff is confusing enough already, let's have our code be clear

backend/test/deal-observer.test.js
NikolasHaimerl: The unique constraint does prevent the insertion of the duplicate. The handling of a conflict on unique constraint handles the conflict. At least that is how I understood it.

juliangruber: Here it _does_ prevent the insertion of a new record, but it doesn't throw an error doing so. I think the comment is correct, but we could add about the error
```suggestion
    // If we only change the id, the unique constraint should prevent the insertion and fail silently
```

juliangruber: @NikolasHaimerl you requested review again without resolving this thread, and there are others that are still unresolved. Could you please first resolve the ones that have already been completed, and address the ones that haven't?

bajtos: Right. I agree the comment correctly describes the unique constraint. I am confused about how the comment relates to what is being tested here.

Suggestion:

```suggestion
    // If we only change the id, the unique constraint which does not include the id will prevent the insertion
    // This test verifies that `storeDeal` handles such situation by ignoring the duplicate deal record
```

backend/test/deal-observer.test.js
bajtos: I created a tech-debt issue to address this: 
- https://github.com/CheckerNetwork/spark-deal-observer/issues/113

backend/test/deal-observer.test.js
juliangruber: +1 to will over should

backend/test/deal-observer.test.js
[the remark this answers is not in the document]
NikolasHaimerl: I renamed it to `actual`

backend/test/deal-observer.test.js
```diff
@@ -107,6 +107,36 @@ describe('deal-observer-backend', () => {
...
+    assert.strictEqual(expected.length, 1)
+    // If we only change the id, the unique constraint which does not include the id should prevent the insertion
+    eventData.id = 2
+    await storeDeal(eventData)
```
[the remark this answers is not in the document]
NikolasHaimerl: I added both variants of tests

backend/test/deal-observer.test.js
NikolasHaimerl: Thanks :)

backend/test/deal-observer.test.js
```diff
@@ -107,6 +107,64 @@ describe('deal-observer-backend', () => {
     await storeBlockEvent(data)
     assert.strictEqual(await countStoredActiveDeals(pgPool), 2n)
   })
+
+  it('handles one client piece stored twice in the same sector and separate sql updates', async () => {
```
juliangruber, line 111: To me this test title isn't clear, does it

"handle one client piece stored twice in the same sector, and also handle separate sql updates"

or does it

"handle one client piece stored twice in the same sector, and stored in separate sql updates"?

juliangruber: I think you want to say that here we test processing two events for one piece, at once. Below, you want to test processing two events for one piece, in succession.

What about these test names instead?

```js
it('simultaneously processes claims for a piece stored twice in the same sector')
```

and

```js
it('serially processes claims for a piece stored twice in the same sector')
```

backend/test/deal-observer.test.js
```diff
@@ -107,6 +107,64 @@ describe('deal-observer-backend', () => {
     await storeBlockEvent(data)
     assert.strictEqual(await countStoredActiveDeals(pgPool), 2n)
   })
+
+  it('simultaneously processes claims for a piece stored twice in the same sector', async () => {
```
bajtos, line 111: In my understanding, "it simultaneously processes" means one query storing two deals, while "it serially processes" means one query per deal. Is that what you had in mind, @juliangruber?

If that's the case, then you need to swap the test titles, @NikolasHaimerl, since the first test is executing two SQL queries.

juliangruber: Yes that's what I had in mind

bajtos (approved):

juliangruber (approved):

PR #114, opened 2025-02-17
chore: count of payload cid states

### LINKED ISSUE #112: Telemetry entry for payload state
We currently only have a chart which tracks the number of payload CIDs that are set to null. We also want to have this number for all possible payload CID states.

NikolasHaimerl:
This PR adds a function `countStoredActiveDealsWithPayloadState` which counts the number of active deals that have a specific payload CID state. 
The different counts are added to telemetry to give a better insight in the current state of payload cid resolving. 
Closes #112


backend/test/look-up-payload-cids.test.js
```diff
@@ -80,6 +80,68 @@ describe('deal-observer-backend look up payload CIDs', () => {
...
+    const getDealPayloadCidCalls = []
+    const makePaylodCidRequest = async (providerId, pieceCid) => {
+      getDealPayloadCidCalls.push({ providerId, pieceCid })
+      const payloadCid = payloadCIDs.get(JSON.stringify({ minerId: providerId, pieceCid }))
+      return payloadCid?.payloadCid
+    }
+
+    await lookUpPayloadCids(makeRpcRequest, makePaylodCidRequest, pgPool, 10000)
+    missingPayloadCids = await countStoredActiveDealsWithPayloadState(pgPool, PayloadRetrievabilityState.Resolved)
+    assert.strictEqual(missingPayloadCids, 251n)
```
bajtos, line 96: Where is the number `251n` coming from? Why it isn't `1n`, `250n` or `1000n` instead? I would like the test case to make that answer obvious.

In my previous comment, I suggested explicitly inserting rows with active deals. When done right, such solution will also provide the answer to the question of how many missing payload CIDs should be reported.

A high-level proposal:


```js
it('counts number of resolved payload CIDs', async () => {
  await givenStoredActiveDeal({pieceCID: 'baga1', payloadRetrievabilityState: PayloadRetrievabilityState.Unresolved})
  await givenStoredActiveDeal({pieceCID: 'baga2', payloadRetrievabilityState: PayloadRetrievabilityState.Unresolved})
  await givenStoredActiveDeal({pieceCID: 'baga3', payloadRetrievabilityState: PayloadRetrievabilityState.Resolved})
  await givenStoredActiveDeal({pieceCID: 'baga4', payloadRetrievabilityState: PayloadRetrievabilityState.TerminallyUnretrievable})

  const actual = await countStoredActiveDealsWithPayloadState(pgPool, PayloadRetrievabilityState.Unresolved)
  assert.strictEqual(actual, 2)
})

async function givenStoredActiveDeal(pgPool, props) {
  const deal = {
    pieceCID: 'baga',
    pieceSize: 32,
    // TODO - provide reasonable default values for all required properties
    ...props
  })
  await pgPool.query(`
    INSERT INTO active_deals
    (piece_cid, piece_size, etc)
    VALUES
    ($1, $2, etc)
  `, [
    deal.pieceCID,
    deal.pieceSize,
    // etc.
  ])
  // ^^^ IMPORTANT: this query must trigger an error when the unique constraint
  // is violated, to tell us when a test accidentally creates duplicate deals
}
```

bajtos (changes requested):
Great idea to add this new telemetry to give us more visibility 👍🏻 

The implementation is correct; let's improve the non-functional aspects now.

backend/test/look-up-payload-cids.test.js
```diff
@@ -80,6 +80,68 @@ describe('deal-observer-backend look up payload CIDs', () => {
     missingPayloadCids = await countStoredActiveDealsWithUnresolvedPayloadCid(pgPool)
     assert.strictEqual(missingPayloadCids, 85n)
   })
+
+  it('piece indexer count number of resolved paylod CIDs', async () => {
```
bajtos, line 84: I would also like to discuss the test naming conventions. The BDD style (`describe/it`) is designed to write test cases as English sentences.

If you prefer to use the XP-style, where the test case name is a name that stands on its own, then please use `test/suite` instead of `it/describe`.

```js
import {test, suite} from 'node:test'

// ...
test('piece indexer count number of resolved paylod CIDs', async () => {
// ...
```

API docs:
- https://nodejs.org/api/test.html#testname-options-fn
- https://nodejs.org/api/test.html#suitename-options-fn

Quoting from https://nodejs.org/api/test.html#describe-and-it-aliases:

> Suites and tests can also be written using the `describe()` and `it()` functions. [describe()](https://nodejs.org/api/test.html#describename-options-fn) is an alias for [suite()](https://nodejs.org/api/test.html#suitename-options-fn), and [it()](https://nodejs.org/api/test.html#itname-options-fn) is an alias for [test()](https://nodejs.org/api/test.html#testname-options-fn).

backend/lib/look-up-payload-cids.js
```diff
@@ -55,6 +55,17 @@ export async function countStoredActiveDealsWithUnresolvedPayloadCid (pgPool) {
   return result.rows[0].count
 }
 
+/**
+ * @param {Queryable} pgPool
+ * @param {Static< typeof PayloadRetrievabilityStateType>} state
+ * @returns {Promise<Array<Static< typeof ActiveDealDbEntry>>>}
```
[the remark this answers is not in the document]
juliangruber: Agreed

backend/test/look-up-payload-cids.test.js
juliangruber: I also mentioned this in a review before :)

backend/bin/deal-observer-backend.js
```diff
@@ -125,10 +125,18 @@ export const lookUpPayloadCidsLoop = async (makeRpcRequest, getDealPayloadCid, p
...
+      const totalNumOfTerminallyUnretrievablePayloadCidStates = await countStoredActiveDealsWithPayloadState(pgPool, 'PAYLOAD_CID_TERMINALLY_UNRETRIEVABLE')
+      const totalNumOfNotYetQueriedPayloadCidStates = await countStoredActiveDealsWithPayloadState(pgPool, 'PAYLOAD_CID_NOT_QUERIED_YET')
       if (INFLUXDB_TOKEN) {
         recordTelemetry('look_up_payload_cids_stats', point => {
-          point.intField('total_unresolved_payload_cids', numOfUnresolvedPayloadCids)
+          point.intField('total_unresolved_payload_cids', totalNumOfUnresolvedPayloadCids)
+          point.intField('total_unresolved_payload_cid_state', totalNumOfUnresolvedPayloadCidStates)
+          point.intField('total_resolved_payload_cid_state', totalNumOfResolvedPayloadCidStates)
+          point.intField('total_terminally_unretrievable_payload_cid_state', totalNumOfTerminallyUnretrievablePayloadCidStates)
+          point.intField('total_not_yet_queried_payload_cid_state', totalNumOfNotYetQueriedPayloadCidStates)
```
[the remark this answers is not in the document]
juliangruber: agreed with Miro. The current version, eg "total_unresolved_payload_cid_state", makes me wonder how you can count "unresolved payload cid state"

juliangruber (changes requested):

backend/test/look-up-payload-cids.test.js
NikolasHaimerl: Let's make this change in a separate PR so we can also change the other parts in the tests where naming is inconsistent with the BDD stile. This way we can move forward with this PR and don't include changes in this PR that are not part of its core functionality.

NikolasHaimerl: I created an issue for this: https://github.com/CheckerNetwork/spark-deal-observer/issues/115

backend/test/look-up-payload-cids.test.js
```diff
@@ -80,6 +80,68 @@ describe('deal-observer-backend look up payload CIDs', () => {
...
+    let missingPayloadCids = await countStoredActiveDealsWithPayloadState(pgPool, PayloadRetrievabilityState.Resolved)
+    assert.strictEqual(missingPayloadCids, 0n)
+    const getDealPayloadCidCalls = []
+    const makePaylodCidRequest = async (providerId, pieceCid) => {
+      getDealPayloadCidCalls.push({ providerId, pieceCid })
+      const payloadCid = payloadCIDs.get(JSON.stringify({ minerId: providerId, pieceCid }))
+      return payloadCid?.payloadCid
+    }
+
+    await lookUpPayloadCids(makeRpcRequest, makePaylodCidRequest, pgPool, 10000)
```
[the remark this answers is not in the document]
juliangruber: ping @NikolasHaimerl

backend/test/look-up-payload-cids.test.js
NikolasHaimerl: I changed the testing to be more explicit

backend/test/look-up-payload-cids.test.js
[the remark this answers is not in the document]
juliangruber: Got it, can you please mention such an then hit "Resolve conversation"?

backend/test/look-up-payload-cids.test.js
juliangruber: as in https://github.com/CheckerNetwork/spark-deal-observer/pull/114#discussion_r1959641941

backend/test/resolve-payload-cids.test.js
```diff
@@ -212,4 +212,92 @@ describe('deal-observer-backend piece indexer payload retrieval', () => {
...
+      term_start_epoch: 1,
+      term_min: 1,
+      term_max: 1,
+      sector_id: 1,
+      payload_cid: undefined,
+      payload_retrievability_state: PayloadRetrievabilityState.NotQueried,
+      last_payload_retrieval_attempt: undefined
+    })
+    await storeActiveDeals([defaultActieDeal], pgPool)
+    await storeActiveDeals([{ ...defaultActieDeal, miner_id: 2, payload_retrievability_state: PayloadRetrievabilityState.Unresolved }], pgPool)
```
bajtos, line 254: What happens when we need to add more deals to the test setup here and we forget that each deal must be unique - e.g. we forget to change `miner_id`.

```js
await storeActiveDeals([{ ...defaultActieDeal, miner_id: 2, payload_retrievability_state: PayloadRetrievabilityState.Unresolved }], pgPool)
await storeActiveDeals([{ ...defaultActieDeal, miner_id: 2, payload_retrievability_state: PayloadRetrievabilityState.TerminallyUnretrievable }], pgPool)
```

Will the new version fail and tell us about the violation of the unique constraint?

Or will it silently ignore the duplicate deal, leading us to believe the code works as intended, which is not true?

bajtos (changes requested):

backend/test/resolve-payload-cids.test.js
NikolasHaimerl: I believe the fix intended was that it would ignore duplicate deals, no?

NikolasHaimerl: I believe the intent of the hot fix that we did last week was that the unique constraint is ignored on duplicate values, no?

bajtos: We discussed this thread yesterday on a Zoom call, it's related to https://github.com/CheckerNetwork/spark-deal-observer/pull/114#discussion_r1958483960

I find it important to change the tests to use a different method for creating test-data deal than `storeActiveDeals`. When inserting test-data deals to the DB, we don't want to ignore conflicts (violations of the unique constraint), we want tests to fail right at the point where the attempt to insert a duplicate deal.

bajtos (changes requested):

juliangruber (dismissed):

backend/bin/deal-observer-backend.js
```diff
@@ -125,11 +125,19 @@ export const resolvePayloadCidsLoop = async (makeRpcRequest, makePayloadCidReque
     const maxDeals = 1000
     try {
       const numOfPayloadCidsResolved = await resolvePayloadCids(makeRpcRequest, makePayloadCidRequest, pgPool, maxDeals)
-      const numOfUnresolvedPayloadCids = await countStoredActiveDealsWithUnresolvedPayloadCid(pgPool)
+      const totalNumOfUnresolvedPayloadCids = await countStoredActiveDealsWithUnresolvedPayloadCid(pgPool)
+      const totalNumOfResolvedPayloadCidStates = await countStoredActiveDealsWithPayloadState(pgPool, 'PAYLOAD_CID_RESOLVED')
```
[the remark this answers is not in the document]
NikolasHaimerl: Addressed in de7ad74

backend/test/resolve-payload-cids.test.js
```diff
@@ -214,7 +229,7 @@ describe('deal-observer-backend piece indexer payload retrieval', () => {
   })
 
   it('piece indexer count number of resolved paylod CIDs', async () => {
-    const defaultActieDeal = Value.Parse(ActiveDealDbEntry, {
+    const defaultActiveDeal = Value.Parse(ActiveDealDbEntry, {
```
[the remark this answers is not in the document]
NikolasHaimerl: Addressed in de7ad74

backend/test/resolve-payload-cids.test.js
NikolasHaimerl: Addressed in de7ad74

backend/test/resolve-payload-cids.test.js
```diff
@@ -108,13 +109,61 @@ describe('deal-observer-backend piece indexer payload retrieval', () => {
...
   const minerPeerId = 'MINER_PEER_ID'
-  const pieceCid = 'PIECE_CID'
   const now = Date.now()
   const fetchMinerId = async () => {
     return { PeerId: minerPeerId }
   }
+  /**
+  * @param {Static<typeof ActiveDeal >[]} activeDeals
+  **/
+  const storeActiveDealsCustom = async (activeDeals) => {
```
bajtos, line 119: The adjective "custom" does not tell us much about how this function differs from the version used in production. Let's find a more descriptive name please.

What information we want to convey:
- The production version handles duplicate entries by silently ignoring conflicts
- The version used in tests throws an error when attempting to insert a duplicate deal

Proposed name for the test:

- `storeUniqueActiveDeals`

I like that this name hints at the person reading the code to check whether the deal fields have unique values.

In other places, I prefix test-data builders with `given`. @pyropy uses the prefix `with`. These are viable options, too.

- `givenActiveDeals`
- `givenUniqueActiveDeals`
- `withActiveDeals`
- `withUniqueActiveDeals`

Which name do you prefer? Can you come up with an even better alternative? Let's discuss.

bajtos (changes requested):
Almost there!

backend/test/resolve-payload-cids.test.js
NikolasHaimerl: I think the name `withUniqueActiveDeals` sounds good and is descriptive. Thanks for the suggestion. 
ae43469

backend/test/resolve-payload-cids.test.js
```diff
@@ -247,4 +274,36 @@ describe('deal-observer-backend piece indexer payload retrieval', () => {
     await resolvePayloadCids(fetchMinerId, resolvePayloadCid, pgPool, 10000, now)
     assert.strictEqual(payloadsCalled, 1)
   })
+
+  it('piece indexer count number of resolved paylod CIDs', async () => {
+    await withUniqueActiveDeals([DEFAULT_ACTIVE_DEAL])
+    await withUniqueActiveDeals([{ ...DEFAULT_ACTIVE_DEAL, miner_id: 2, payload_cid: 'PAYLOAD_CID', payload_retrievability_state: PayloadRetrievabilityState.Resolved }])
+    await withUniqueActiveDeals([{ ...DEFAULT_ACTIVE_DEAL, miner_id: 3, payload_cid: 'PAYLOAD_CID', payload_retrievability_state: PayloadRetrievabilityState.Resolved }])
```
bajtos, line 281: I am curious - why do we have the complex implementation in `withUniqueActiveDeals` to support inserting multiple deals in a single SQL query when the actual tests insert deals one by one?

In my experience, simpler is better, so I propose to change `withUniqueActiveDeals` to `withUniqueActiveDeal` and insert a single deal only.

Alternatively, let's rework the tests to leverage the full power of `withUniqueActiveDeals`.

```js
    await withUniqueActiveDeals([
      DEFAULT_ACTIVE_DEAL,
      { ...DEFAULT_ACTIVE_DEAL, miner_id: 2, payload_cid: 'PAYLOAD_CID', payload_retrievability_state: PayloadRetrievabilityState.Resolved },
      { ...DEFAULT_ACTIVE_DEAL, miner_id: 3, payload_cid: 'PAYLOAD_CID', payload_retrievability_state: PayloadRetrievabilityState.Resolved }
  ])
```

Feel free to address this comment in a follow-up pull request if you prefer. It's not blocking this PR from being merged.

bajtos (approved):

backend/test/resolve-payload-cids.test.js
NikolasHaimerl: 711af9d

Issue #115, opened 2025-02-18
Consistend BDD stile test naming


NikolasHaimerl:
I would also like to discuss the test naming conventions. The BDD style (`describe/it`) is designed to write test cases as English sentences.

If you prefer to use the XP-style, where the test case name is a name that stands on its own, then please use `test/suite` instead of `it/describe`.

```js
import {test, suite} from 'node:test'

// ...
test('piece indexer count number of resolved paylod CIDs', async () => {
// ...
```

API docs:
- https://nodejs.org/api/test.html#testname-options-fn
- https://nodejs.org/api/test.html#suitename-options-fn

Quoting from https://nodejs.org/api/test.html#describe-and-it-aliases:

> Suites and tests can also be written using the `describe()` and `it()` functions. [describe()](https://nodejs.org/api/test.html#describename-options-fn) is an alias for [suite()](https://nodejs.org/api/test.html#suitename-options-fn), and [it()](https://nodejs.org/api/test.html#itname-options-fn) is an alias for [test()](https://nodejs.org/api/test.html#testname-options-fn).

_Originally posted by @bajtos in https://github.com/CheckerNetwork/spark-deal-observer/pull/114#discussion_r1958490551_

PR #154, opened 2025-04-14
chore: applied BDD style to tests

NikolasHaimerl:
Closes https://github.com/CheckerNetwork/spark-deal-observer/issues/115


bajtos (approved):
I assumed this pull request is only renaming tests and moving code around, so I only quickly skimmed through the changes.

I don't see any obvious problem, LGTM 👍🏻 

Thank you for improving our test suite!

